PR #1, opened 2016-03-16
Review: Initial script to create driver file

GitEdvard:
The core method is create_driver_file in perform_dilution.py.  The method creates a start for a Hamilton driver file for dilution. Some columns are still hardcoded values. There is also a Hamilton driver file reader in order to perform tests of the script outcome. The tests on the script connects to db.


DoTests.bat
```diff
@@ -0,0 +1,2 @@
+py -2 -m clarity_ext.automated_tests.test_cases.dilute_tests.hamilton_tests2
```
withrocks: This script we would probably change to using a python test runner, such as nosetests, and create both .sh and .bat for executing it.

clarity_ext/automated_tests/test_cases/dilute_file_reader/hamilton_driver_file_tests.py
```diff
@@ -0,0 +1,116 @@
+# Tests for reading from a Hamilton driver file,
```
withrocks: We will probably eventually move the tests to the root directory.

clarity_ext/automated_tests/test_cases/dilute_file_reader/hamilton_driver_file_tests.py
```diff
@@ -0,0 +1,116 @@
...
+# using the utility dilute_filer_reader
+
+import unittest
+import os
+import inspect
+from ...utility.dilute_filer_reader.hamilton_driver_file_reader import HamiltonReader as FileReader
```
withrocks: We used relative paths before in the arteria project and figured that they are not recommended by the community (or the core team, don't remember the details). This can be fixed when we update the test structure.

clarity_ext/automated_tests/test_cases/dilute_file_reader/hamilton_driver_file_tests.py
```diff
@@ -0,0 +1,116 @@
...
+        self.assertEqual(int(contents), 16,
+                         "Target well pos reference not right")
+
+    def test_reference_target_plate_pos(self):
+        contents = self.file_reader.dict_matrix["SX614_T11.v1"][self.column_ref.target_plate_pos]
+        self.assertEqual(contents, "END1",
+                         "Target plate pos reference not right")
+
+if __name__ == "__main__":
+    unittest.main()
```
withrocks: Cool. Great to have tests from day 1.

clarity_ext/automated_tests/test_cases/dilute_tests/hamilton_tests.py
```diff
@@ -0,0 +1,52 @@
...
+SAMPLE1 = "EdvardsProv1Updated"
+SAMPLE2 = "EdvardsProv3"
+SAMPLE3 = "EdvardsProv2"
+SAMPLE4 = "EdvardsProv4"
+
+
+class HamiltonTests(unittest.TestCase):
+
+    def setUp(self):
+        driver_file_contents = perform_dilution.create_driver_file(TEST_PROCESS_URI, None)
```
withrocks: We can look into if the framework stuff in the clarity-ext project wouldn't help with this.

It provides a different take on integration tests, where the tests are executed once and then again on each build, comparing the results as a whole. They have the extra advantage that the process step does not need to exist anymore, since they use cached requests/responses.

clarity_ext/dilute_epp/perform_dilution.py
```diff
@@ -0,0 +1,164 @@
+# Calculates dilute volumes, volume from sample and volume from buffer
```
withrocks: This script could be refactored as an extension, that can be executed with the command `clarity-ext extension ...`, where the boilerplate could would go down to some base class that uses a context object. I will add some notes on that in the following comments.

clarity_ext/dilute_epp/perform_dilution.py
```diff
@@ -0,0 +1,164 @@
...
+# for each analyte within a given process.
+# User provide target concentration and target volume for each analyte.
+# Give error if any volume is less than 2 ul
+# Give error if any volume exceeds 50 ul
+
+from genologics.lims import *
```
withrocks: ... imports wouldn't need `argparse` or `genologics` - only higher level dependencies.

withrocks:
Looks cool. I think I'll look into how this can be written as an extension (what changes I'll need to make) and then I can demo that afterwards.

PR #3, opened 2016-04-04
Ee lims230 dilution with happy path 2

GitEdvard:
Create a robot driver file for dilutions.
Restrictions:
* Only generates the driver file, do not update fields in db, like volume or concentration
* Do not generate the dilute file
* Have only been tested on one process, constructed for this purpose
* The script will not work if the input analytes lack concentration udf


withrocks:
I will add comments and then make corresponding changes myself..

.idea/runConfigurations/create_hamilton_dilution___test.xml
```diff
@@ -4,6 +4,7 @@
     <option name="PARENT_ENVS" value="true" />
     <envs>
       <env name="PYTHONUNBUFFERED" value="1" />
+      <env name="PYTHONPATH" value="C:/GitRoot/clarity-ext" />
```
withrocks: Should be unnecessary, will be removed.

clarity_ext/extension_context.py
```diff
@@ -156,6 +144,48 @@ def cleanup(self):
...
 
+class MatchedAnalytes:
+    """ Provides a set of  matched input - output analytes for a process.
+    When fetching these by the batch_get(), they come in random order
+    """
+    def __init__(self, input_analytes, current_step, advanced):
+        self._input_analytes = input_analytes
+        self.advanced = advanced
+        self.current_step = current_step
+        (matched_in_analytes, matched_out_analytes) = self._match_analytes()
```
withrocks: Parentheses are not necessary.

clarity_ext/extension_context.py
```diff
@@ -156,6 +144,48 @@ def cleanup(self):
...
+    def _output_analytes(self):
+        (analytes, info) = self.current_step.analytes()
+        if not info == 'Output':
+            raise ValueError("No output analytes for this step!")
+        resources = self.advanced.lims.get_batch(analytes)
+        return [Analyte(resource) for resource in resources]
+
+    def _match_analytes(self):
+        """ Match input and output analytes with sample ids"""
+        input_dict = dict([(input_.sample.id, input_)
```
withrocks: It's more readable to use a dictionary comprehension.

clarity_ext/extension_context.py
```diff
@@ -156,6 +144,48 @@ def cleanup(self):
...
+                           for input_ in self._input_analytes])
+        matched_analytes = [(input_dict[output_.sample.id], output_)
```
withrocks: I'd prefer an underscore prefix rather than a postfix.

clarity_ext/dilution.py
```diff
@@ -0,0 +1,151 @@
...
+        plate_base_number = self.plate_size_y * self.plate_size_x + 1
+        plate_sorting = self.source_plate_sorting_map[
+            dilute.source_container.id]
+        # Sort order for wells are always based on down first indexing
+        # regardless the robot type
+        well_index = self._calculate_well_index_down_first(dilute.source_well)
+        return plate_sorting * plate_base_number + well_index
+
+    def _calculate_well_index_down_first(self, well):
+        (y, x) = well.get_coordinates()
```
withrocks: I think this can easily confusing, e.g. seeing the x,y coordinates in this order (I'm more used to matrix convention in that case, i.e. `i`,`j` instead of `x`,`y`). I suggest that we replace this with a named tuple called `PlatePosition` that has the attributes `row` and `col`.

clarity_ext/dilution.py
```diff
@@ -0,0 +1,151 @@
...
+        return x * self.plate_size_y + y + 1
```
withrocks: Same goes with the size element here. I suggest refactoring this to `self.plate.size.width`.

clarity_ext/dilution.py
```diff
@@ -0,0 +1,151 @@
...
+    def find_sort_number(self, dilute):
+        """Sort dilutes according to plate and well positions in source
+        :param dilute:
+        """
+        plate_base_number = self.plate_size_y * self.plate_size_x + 1
+        plate_sorting = self.source_plate_sorting_map[
+            dilute.source_container.id]
+        # Sort order for wells are always based on down first indexing
+        # regardless the robot type
+        well_index = self._calculate_well_index_down_first(dilute.source_well)
```
withrocks: I suggest that we move the index calculation down to the well, having it know which plate it belongs to, and thus it will know this.

clarity_ext/extension_context.py
```diff
@@ -156,6 +144,48 @@ def cleanup(self):
                 os.remove(path)
 
 
+class MatchedAnalytes:
```
withrocks: It seems that it's not necessary to have a class for encapsulating this. I would rather just create helper methods/functions to use in the context object.

withrocks:
Looks good now, I'll merge this.

PR #4, opened 2016-04-08
Implement full integration tests

withrocks:
- All integration tests can run through ./validate.sh
- Context "artifacts" (i.e. downloaded files) are now also cached
- Remove unused code


withrocks:
I'm merging this myself since we're still not in production.

PR #5, opened 2016-04-14
Implement support for generic containers

withrocks:
- Context can return input_containers, which in turn have all samples on
  them, so that should be used instead of the input/output map (not
  supporting output containers yet)
- Clean up plate positioning code. Wells should now always use the
  PlatePosition namedtuple, which in turn can represent themselves in
  different formats (e.g. A:1 or 1:1)
- Use the Container class (more generic) instead of the Plate class
- Support Strip Tubes
- Minor: Update run configurations


withrocks:
Still in staging only. Merging directly.

PR #7, opened 2016-05-03
Support Windows line endings

withrocks:


GitEdvard:
Question mark no 1: All lab staff is working on windows machines, shouldn't the windows line endings "\r\n"be the default? Now, only the fragment analyser script is using the windows line ending. 

Question mark no 2: I'm confused about why the line endings in my dilution driver files works on windows machines with the current version of Clarity-ext. The dilution driver files are built up with "\n". I just sent a driver file to Ingvar, and it looks OK on his machine as well. 

These are minor objections though, so ping me as soon as you want me to merge!

withrocks:
1: Valid point. Since these files are generally for devices rather than client machines, I felt that it was a sensible default to use *nix file endings, although using abstract methods (no default) would perhaps be better.
I think though that the requirements of the devices should be considered rather than the requirements of the clients, since the devices might actually parse these incorrectly if not correctly specified, but all the user has to do is to open this in a better application than notepad - which I suggest they do anyway. (Having said that, perhaps most of the devices do also require Windows.)

Let me know if you think differently.

2: Could it be that you created the driver file on your own machine? When writing to files, Python does by default change line endings according to the platform (except if specifying the `b`inary flag). So if you executed this from your Windows machine, it would have had different results than if running from the server.

PR #9, opened 2016-05-11
Remove header and well pos from FA file

withrocks:


withrocks:
Changes nothing but the FA input file, which is still not in production. Merging after a review by the lab.

PR #10, opened 2016-05-13
Remove clarity_ext_scripts

withrocks:
- Remove the clarity_ext_scripts module. It will be maintained in another repo.


GitEdvard:
Ok

PR #11, opened 2016-05-17
Support for reading XML files as Python objects and related refactoring

withrocks:
- The library now supports reading a particular XML file like a Python object
```
xml = self.context.local_shared_xml("Result XML File")
xml.Element.ChildElement
```
- This requires `lxml` which requires several dependencies on the machine to build. It may be worth it to install with conda (a script for doing that is provided in the private implementation specific repo)
- A class for converting between units
- A class for working with `ResultFile`s
- Necessary refactoring for testability

### DIFF
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -6,6 +6,9 @@ from clarity_ext.dilution import *
 import re
 import shutil
 import clarity_ext.utils as utils
+from lxml import objectify
+from clarity_ext.units import Units
+from clarity_ext.result_file import ResultFile
 
 
 class ExtensionContext:
@@ -26,9 +29,14 @@ class ExtensionContext:
         self._local_shared_files = []
         self.cache = cache
 
-    def local_shared_file(self, file_name):
+        self.units = Units(self.logger)
+        self._update_queue = []
+        self.current_step = clarity_svc.current_step
+
+    def local_shared_file(self, file_name, mode='r'):
         """
-        Downloads the local shared file and returns the path to it on the file system.
+        Downloads the local shared file and returns an open file-like object.
+
         If the file already exists, it will not be downloaded again.
 
         Details:
@@ -81,7 +89,7 @@ class ExtensionContext:
         if local_path not in self._local_shared_files:
             self._local_shared_files.append(local_path)
 
-        return local_path
+        return open(local_path, mode)
 
     def _get_input_analytes(self, plate):
         # Get an unique set of input analytes
@@ -156,6 +164,14 @@ class ExtensionContext:
                                  "that it won't be uploaded again".format(path))
                 # TODO: Handle exception
                 os.remove(path)
+
+    def local_shared_xml(self, name):
+        """
+        Returns a local copy of the xml file as a Python object
+        """
+        with self.local_shared_file(name, "r") as fs:
+            tree = objectify.parse(fs)
+            return tree.getroot()
 
 
 class MatchedAnalytes:
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -9,6 +9,7 @@ import clarity_ext.utils as utils
 from lxml import objectify
 from clarity_ext import UnitConversion
 from clarity_ext.result_file import ResultFile
+from clarity_ext.utils import lazyprop
 
 
 class ExtensionContext:
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -7,7 +7,7 @@ import re
 import shutil
 import clarity_ext.utils as utils
 from lxml import objectify
-from clarity_ext.units import Units
+from clarity_ext import UnitConversion
 from clarity_ext.result_file import ResultFile
 
 
@@ -29,7 +29,7 @@ class ExtensionContext:
         self._local_shared_files = []
         self.cache = cache
 
-        self.units = Units(self.logger)
+        self.units = UnitConversion(self.logger)
         self._update_queue = []
         self.current_step = clarity_svc.current_step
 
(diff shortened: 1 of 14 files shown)


clarity_ext/units.py
```diff
@@ -0,0 +1,30 @@
+import math
+import logging
+
```
GitEdvard: I think small classes like this should find it's home in some common module, like "misc".

test/unit/clarity_ext/test_units.py
```diff
@@ -0,0 +1,25 @@
+import unittest
+from clarity_ext.units import Units
+
```
GitEdvard: The name 'units' is slightly confusing with unit tests. maybe si_units, si_prefix?

clarity_ext/result_file.py
```diff
@@ -0,0 +1,14 @@
+class ResultFile:
```
GitEdvard: Shouldn't this class be placed in the domain module? It's a very small class to be contained in one module.

clarity_ext/context.py
```diff
@@ -1,5 +1,4 @@
 from genologics.config import BASEURI, USERNAME, PASSWORD
```
GitEdvard: Ok

clarity_ext/extensions.py
```diff
@@ -107,6 +107,9 @@ def execute(self, module, mode, run_arguments_list=None, config=None, artifacts_
                     module, self.RUN_MODE_FREEZE))
```
GitEdvard: Ok

clarity_ext/pdf.py
```diff
@@ -2,8 +2,8 @@
```
GitEdvard: This is another very small file. It not placing it in a common module we could place it in a subdirectory?

clarity_ext/context.py
```diff
@@ -173,6 +173,27 @@ def local_shared_xml(self, name):
             tree = objectify.parse(fs)
             return tree.getroot()
```
GitEdvard: Ok

clarity_ext/context.py
```diff
@@ -173,6 +173,27 @@ def local_shared_xml(self, name):
...
+                yield output["uri"]
+
+    def update(self, obj):
+        """Add an object that has a commit method to the list of objects to update"""
+        self._update_queue.append(obj)
+
+    def commit(self):
+        """Commits all objects that have been added via the update method, using batch processing if possible"""
+        # TODO: Implement batch processing
+        for obj in self._update_queue:
```
GitEdvard: Ok, implementing the batch processing is fairly easy, so I suggest to do this soon

setup.py
```diff
@@ -4,7 +4,8 @@
 from setuptools import find_packages, setup
```
GitEdvard: Ok

clarity_ext/units.py
withrocks: I think it might be valid to have it in a package called e.g. `utils` or `misc` - but I think the modules here are quickly getting really large, e.g. the domain module. That's very common in Python projects, but personally I think it's a bad habit, since it makes getting an overview of the project difficult.

Taking a successful Python library as an example, Luigi (https://github.com/spotify/luigi/tree/master/luigi), they often have small modules. They then import these modules to the package level by importing them in `__init__.py`. See [here](https://github.com/spotify/luigi/blob/master/luigi/task_status.py) and [here](https://github.com/spotify/luigi/blob/master/luigi/__init__.py).

In this case, I believe that this class will get ~1 page long when finished, and it contains a clearly defined thing (something that handles unit conversion), so I think it should be in its own module. I'm not sure I see the advantage of moving it into another package though - since it's unclear why this is `misc`, but not for example `pdf.py`.

test/unit/clarity_ext/test_units.py
withrocks: Good point. I've renamed it to `UnitConversion` and am now importing it on the root package level, so it can be used with `from clarity_ext import UnitConversion`.

clarity_ext/result_file.py
withrocks: The idea is that the class will be larger with time. I think the domain module is starting to get too large, containing stuff that really doesn't belong together.

This was intended as the start of refactoring the domain module to a package, in order to increase maintainability. I'll use this opportunity to do that.

clarity_ext/pdf.py
withrocks: Regarding this, I feel that it's most important that developers can quickly find their way through the project - without running a search or viewing the project in a compiled class representation in some IDE.

Even if it doesn't contain much at the moment, the module encapsulates entirely what it's there for. So if someone needs to solve a bug regarding pdfs, he can assume that he only has to read through this small module.

The problem I have with packages as directories is that it's not very clear what should go where. I think a module that's put in a directory where it doesn't really belong is better off in the root. PDF parsing is not more miscellaneous than e.g. unit_conversion, so putting them into e.g. something called `misc` begs the question where to draw that line.

I think it might make sense to move this into a package called e.g. `io`. But currently there is no other file that should clearly go into that package, so I suggest we wait with that.

withrocks: ... sorry for the essay.

clarity_ext/context.py
withrocks: I agree, it's simple, but I wanted to limit the scope of the things needing tests here (as well as the scope). I've added a ticket so that this won't be forgotten.

withrocks:
Thanks for the comments. You can merge this if there are no further comments.

GitEdvard:
Nice

PR #13, opened 2016-05-27
Create an entity based mock

GitEdvard:
Overwrite methods in genologics.entity derived classes in order to use fake data.


withrocks:
There is something funky going on here with my changes going in as separate commits. We'll need to fix that together and create a new pull request.

withrocks:
I recommend fixing this in this way:
* Create a new branch from `develop`, call it e.g. `ee-LIMS268-Test-env-clean`. This branch is now entirely clean.
* Find out which commits should actually go in, e.g. by diffing the results of `git-log --oneline`. In this case the commits are:
  - **d8a460c**
  - aefb318: Ignored, no change from commit before
  - 8b4a93d: Ignored, merges from develop are not needed
  - **4efa3c7**
  - **a5d577b**
* Merge these into the clean branch by using git `cherry-pick`. You now have new commits after latest on `develop`.
* Diff this against the original branch (ee-LIMS269-Test-env-clean), it should show no diff.
* Create a new pull request.
* Profit.

PR #19, opened 2016-05-30
Create entity based mock

GitEdvard:
Updating functions in genologics.entities in order to make a test environemt reading locally only. Core class is FakingEntityMonkey.

### DIFF
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -1,5 +1,6 @@
 from genologics.config import BASEURI, USERNAME, PASSWORD
 from genologics.entities import *
+from genologics.lims import Lims
 import requests
 import os
 from clarity_ext.dilution import *
@@ -29,7 +30,6 @@ class ExtensionContext:
         self.logger = logger or logging.getLogger(__name__)
         self._local_shared_files = []
         self.cache = cache
-
         self.units = UnitConversion(self.logger)
         self._update_queue = []
         self.current_step = clarity_svc.current_step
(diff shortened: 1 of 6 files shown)


clarity_ext/context.py
```diff
@@ -196,6 +196,104 @@ def commit(self):
             obj.commit()
 
 
+class FakingEntityMonkey:
+    """
+    Prepare genologics.entities for an entity based mock by
```
withrocks: The docstring could be clearer. How should it be used and what's the expected outcome.

clarity_ext/context.py
```diff
@@ -196,6 +196,104 @@ def commit(self):
...
+    def __init__(self):
+
+        self._overwrite_entity()
+        self._overwrite_string_descriptor()
+        self._overwrite_process()
+        self._overwrite_lims()
+        self._overwrite_entity_list_descriptor()
+
+    def reset(self):
+        EntityListDescriptor.__get__ = self.entity_list_descriptor_get
```
withrocks: We talked about the approach of using mock objects defined only in the test itself instead, i.e. one abstraction layer closer, so to speak. Those kinds of tests tend to be super fast and probably simpler to maintain. Are you not considering doing that anymore? If you do, and this is temporary, I would suggest adding a comment on that, and again, move it closer to the test so that it's not something one needs to read to get an overview of the project.

test/unit/clarity_ext/dilution/hamilton_tests.py
```diff
@@ -17,6 +20,9 @@ class HamiltonTests(unittest.TestCase):
     """
 
     def setUp(self):
+        self.mock = DilutionMock1()
```
withrocks: There are a few things I would change here:
 * Skip that helper class called `FakingEntityMonkey`
 * Move all mock logic from `DilutionMock1` into the test itself - possibly dragging in the mock library. My view of tests is that they should generally be the only thing one needs to read in order to understand a unit.
 * Mock the methods with either monkey patching or by using mock objects from the mock library.

This would get rid of several classes as well as increasing readability (perhaps arguably).
 
I'll see if I can sketch my ideas in another branch real quick and see what you think.

clarity_ext/context.py
```diff
@@ -196,6 +196,104 @@ def commit(self):
             obj.commit()
 
 
+class FakingEntityMonkey:
```
GitEdvard: Using the word monkey in a class that is solely doing monkey patching couldn't be that bad or?
and, I forgot to move the class...

clarity_ext/context.py
GitEdvard: The purpose of this class is to be able to crate an entity based mock. This is not the mock itself

test/unit/clarity_ext/dilution/hamilton_tests.py
GitEdvard: No! it's not possible to create the mock without FakingEntityMonkey. And, I'm not fond of putting my mock repository of ~100 lines in my unit test class. I dont see the point in 'get rid of several classes' if there is a clear separation of responsibilities. In addition, there is a base class to the DilutionMock1, that I intend to use later to avoid boilerplate code.

withrocks: I have another branch that I would like to show you that shows another way of doing this. It doesn't require 100 lines of code ;)

clarity_ext/context.py
withrocks: What I'm thinking is that "monkey patching" is similar to "overriding". I would be confused if I saw a class called `EntityOverrider`. I think the technology used to achieve the result is not the main thing. But I would like to show you if I may how I would approach this, in a different branch.

test/unit/clarity_ext/dilution/hamilton_tests.py
withrocks: I agree that it makes sense to move this into another class when using this approach, but I'd like to show another approach. When that approach is used, it starts becoming questionable if it needs to be another class or not. It might be (making the test less readable, but more DRY) but it might also not be.

PR #20, opened 2016-06-08
Use new-style classes

withrocks:
- All classes in clarity_ext should be new-style (inherit from object).
  This is the default in py3.


b97pla:
:+1:

PR #21, opened 2016-06-08
Rename ClarityService to ClaritySession and move it

withrocks:
- The ClarityService class really encapsulates a session. Name it
  accordingly

### DIFF
--- a/clarity_ext/clarity.py
+++ b/clarity_ext/clarity.py
@@ -1,0 +1,18 @@
+from genologics.lims import Lims
+from genologics.config import BASEURI, USERNAME, PASSWORD
+from genologics.entities import Process
+
+
+class ClaritySession(object):
+    """
+    A wrapper around connections to Clarity.
+    """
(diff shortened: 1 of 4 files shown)


clarity_ext/clarity.py
```diff
@@ -0,0 +1,18 @@
+from genologics.lims import Lims
+from genologics.config import BASEURI, USERNAME, PASSWORD
+from genologics.entities import Process
+
+
+class ClaritySession(object):
+    """
+    A wrapper around connections to Clarity.
+    """
```
GitEdvard: I think it's confusing not to distinguish between current_step (instance) and current_step (id). I would prefer current_step_id as input argument.

GitEdvard:
Ok, let me know when I should merge!

clarity_ext/clarity.py
withrocks: I agree - I'll fix that.

PR #22, opened 2016-06-09
Add a repository for analytes in the current step

withrocks:
- Fetch the `Analyte` domain objects through a repository only.
  Don't access it directly through the REST API. This makes it possible
  to mock this repository only when unit testing classes that use it.
- Stop sending in a `Plate` object
- Remove the MatchedAnalytes class, replace it with sorting logic in the
  repository
- Update the dilutes module in relation to this
- Use the Container rather than the Plate object.
- Provide a resource when creating the domain `Container` object.

### DIFF
--- a/clarity_ext/dilution.py
+++ b/clarity_ext/dilution.py
@@ -1,4 +1,4 @@
-from domain import *
+from clarity_ext.domain.validation import ValidationException, ValidationType
 
 
 class Dilute(object):
@@ -54,7 +54,6 @@ class RobotDeckPositioner(object):
         # regardless the robot type
         return plate_sorting * plate_base_number + dilute.source_well.index_down_first
 
-
     @staticmethod
     def _build_plate_position_map(plate_sorting_map, plate_pos_prefix):
         # Fetch an unique list of container names from input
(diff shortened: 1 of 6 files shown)


clarity_ext/dilution.py
```diff
@@ -79,27 +78,38 @@ def _build_plate_sorting_map(containers):
         return plate_position_numbers
 
     def __str__(self):
-        return "<{type} {robot} {rows}x{cols}>".format(type=self.__class__.__name__,
-                                                       robot=self.robot_name,
-                                                       rows=self.plate_size_y,
-                                                       cols=self.plate_size_x)
-
-PLATE_TYPE_96_WELL = 1
```
GitEdvard: I prefer the rows and cols notation before height and width. This is also a notation that the lab-people use for plates and tube racks. I think however that the size_y and size_x should be removed.

clarity_ext/dilution.py
```diff
@@ -79,27 +78,38 @@ def _build_plate_sorting_map(containers):
...
+        Input and output analytes must be in the same order.
+        """
+        input_analytes, output_analytes = step_input_output_repo.all_analytes()
+        assert len(input_analytes) == len(output_analytes), "There must be the same number of input and output analytes"
 
-        self.dilutes = self._init_dilutes(matched_analytes)
-        self.robot_deck_positioner = RobotDeckPositioner(
-            robot_name, self.dilutes, plate)
+        # TODO: Is it safe to just check for the container for the first output analyte?
+        container = output_analytes[0].container
```
GitEdvard: Nope, this has to be updated later to handle more than one output container

clarity_ext/dilution.py
withrocks: That sounds good. To me it doesn't matter if it's row/column or height/width as long as it's clear what it means - but x/y became rather confusing. Using what the lab is most used to is preferable.

Since it's not entirely related to this pull request (I'm just using the plate object rather than the property), I've added a ticket on this.

clarity_ext/dilution.py
withrocks: OK. I've added a ticket on it.

test/integration/domain/test_artifact.py
```diff
@@ -0,0 +1,55 @@
+import unittest
+from clarity_ext.domain.artifact import StepInputOutputMapRepository
+from clarity_ext.clarity import ClaritySession
+import itertools
+
+
```
GitEdvard: In future, we could maybe have a separate database for integration tests? In the long run it will be hard to keep track of which of the steps that shouldn't be updated through the ui and not.

clarity_ext/domain/artifact.py
```diff
@@ -0,0 +1,49 @@
+from clarity_ext.domain import Analyte
+from clarity_ext.domain import Container
+
+
```
GitEdvard: Maybe just StepRepository? Or StepArtifactRepostory? The good thing with the current name is that I know exactly what it is as long as I have druggled though the Clarity XML replresentation, But without a deep dive into the XML the 'InputOutputMap' says not very much at all.

test/integration/domain/test_artifact.py
```diff
@@ -0,0 +1,55 @@
...
+class TestIntegrationAnalyteRepository(unittest.TestCase):
+    """
+    Tests that validate if the analyte repository works as expected, giving confidence in
```
GitEdvard: First sentence in comment is not clear. Skip 'Tests that'.

clarity_ext/domain/artifact.py
```diff
@@ -0,0 +1,49 @@
...
+class StepInputOutputMapRepository(object):
+    """
+    The REST API provides `Artifact`s. These are wrapped as other domain objects.
```
GitEdvard: Nice to have this repo class!

clarity_ext/domain/analyte.py
```diff
@@ -34,11 +37,7 @@ def target_volume(self):
     @property
     def well(self):
         pos = PlatePosition.create(self.resource.location[1])
-        return Well(pos, self.plate)
-
-    @property
-    def container(self):
-        return self.resource.location[0]
```
GitEdvard: We have to rewrite this function later when we include tubes in our scripts as well. But maybe the proper name would be 'tube' then

GitEdvard:
Resources are accessed through the genologics library, not the REST API. I was a bit confused about that.

test/integration/domain/test_artifact.py
withrocks: Yes I agree, this is very much a hack as it's now. It would probably be easiest to maintain if the test would be set up beforehand by api calls.

But then again, it might be easier to maintain only the selenium tests when they have been set up.

test/integration/domain/test_artifact.py
withrocks: Right!

clarity_ext/domain/artifact.py
withrocks: I think you're right about that. It's valid to hide the fact that there's an input/output map behind this - user's don't really need to know that. It's also good to shorten the name.

I'll change it to `StepRepository` with the intent that all data available in a step can be fetched through it in the future. We could change it to `StepArtifactRepository` if the class starts being too large.

clarity_ext/domain/analyte.py
withrocks: The container is still available on the object, it's being sent into the constructor instead.

withrocks:
Right. I see how that can be confusing - I assume that the reason for why I phrased it like that is that to me the genologics library is just a proxy for the REST API. I.e. it only simplifies fetching the resources without any bells and whistles.

GitEdvard:
Ok

PR #23, opened 2016-06-14
Unit tests for the DilutionScheme

withrocks:
Updates the unit tests (available in a non-merged branch) for `DilutionScheme`, so that they use the new `StepRepository`, making the intention of the tests focused on the DilutionScheme logic only.

Includes some cleanup for related previous commits.


clarity_ext/dilution.py
```diff
@@ -72,7 +82,7 @@ def _build_plate_sorting_map(containers):
         # Fetch an unique list of container names from input
         # Make a dictionary with container names and plate position sort numbers
         unique_containers = sorted(list(
-            {container.id for container in containers}))
```
GitEdvard: I don't understand this change. container.id is defined within the container class as container.resource.id

clarity_ext/domain/container.py
```diff
@@ -91,18 +91,25 @@ def __init__(self, mapping=None, container_type=None):
...
+
     @classmethod
     def create_from_rest_resource(cls, resource, artifacts):
         """
         Creates a container based on a resource from the REST API.
         """
-        # TODO: This needs to be cleaned up, i.e. how exactly the rest resource and domain objects should interact
         if resource.type.name == "96 well plate":
             ret = Container(container_type=Container.CONTAINER_TYPE_96_WELLS_PLATE)
+            ret.size = PlateSize(width=resource.type.x_dimension["size"], height=resource.type.y_dimension["size"])
```
GitEdvard: Side note: It looks like sthlm genologics has the notion of x_dimension and y_dimension. If I've misunderstood this, so that it's our code using this notion, we should change it.

clarity_ext/dilution.py
withrocks: I should be `container.id`, not `container.resource.id`. But this is an outdated diff, if you look at the discussion thread in Github, it says "GitEdvard commented on an outdated diff 7 minutes ago". I believe that this is happening because you're looking at individual commits. It might be better to look at "Files changed" instead. (I personally only review each commit in turn if the pull request is too large.)

clarity_ext/domain/container.py
withrocks: Yes, you're right. The REST client and the actual REST XML it's proxying, both use x/y instead of width/height or row/column. So I only use it in code that's parsing it to something less likely to be misunderstood.

clarity_ext/dilution.py
GitEdvard: ok

clarity_ext/domain/artifact.py
```diff
@@ -46,6 +53,21 @@ def _wrap_artifacts(self, artifacts):
         """
         for artifact in artifacts:
             if artifact.type == "Analyte":
-                yield Analyte.create_from_rest_resource(artifact)
+                yield Analyte.create_from_rest_resource(artifact, self.udf_map)
             else:
                 yield artifact
+
+"""
```
GitEdvard: Ok, make clarity-ext independent of our specific set of udfs. Good

test/unit/clarity_ext/dilution/test_dilutions.py
```diff
@@ -0,0 +1,74 @@
...
+        # Test:
+        actual = [
+            [dilute.sample_name,
+             "{}".format(dilute.source_well_index),
+             dilute.source_plate_pos,
+             "{:.1f}".format(dilute.sample_volume),
+             "{:.1f}".format(dilute.buffer_volume),
+             "{}".format(dilute.target_well_index),
+             dilute.target_plate_pos] for dilute in dilution_scheme.dilutes
+        ]
```
GitEdvard: A reflexion on this setup: We copy code from snpseq in order to test it. But, the code in snpseq is there because we want it to be private, or code that should be specific to our place? Here we import that code to the public repo anyway. It seems that in the end we only got a code duplication.

test/unit/clarity_ext/dilution/test_dilutions.py
```diff
@@ -0,0 +1,74 @@
...
+        repo = MagicMock()
+        repo.all_analytes = all_analyte_test_set1
+        dilution_scheme = DilutionScheme(repo, "Hamilton")
+
+        expected = [
+            ['art-name1', '36', 'DNA1', '14.9', '5.1', '34', 'END1'],
+            ['art-name2', '33', 'DNA2', '14.9', '5.1', '17', 'END2'],
+            ['art-name3', '50', 'DNA2', '14.9', '5.1', '44', 'END1'],
+            ['art-name4', '93', 'DNA2', '14.9', '5.1', '69', 'END2']]
+
```
GitEdvard: When developing, I think I would prefer to create a bunch of small little tests, in order to do incremental improvement  of the code according to TDD. I'm thinking a little about how that approach would look like when the feature is finished. Maybe it's more pretty to have one single test to more easily get a grasp of the meaning of the test? Another aspect is when, in future someone breaks something, many small tests will in a more explicit way tell what's wrong.

test/unit/clarity_ext/dilution/test_dilutions.py
```diff
@@ -0,0 +1,74 @@
...
+
+def all_analyte_test_set1():
+    """
+    Returns a list of (inputs, outputs) fake analytes for a particular step.
+
+    Analytes have been sorted, as they would be when queried from the repository.
+    """
+    inputs = [
+        fake_analyte("cont-id1", "art-id1", "sample1", "art-name1", "D:5", concentration=134),
```
GitEdvard: I have some thoughts about the repository pattern implemented previously, and the matching functionality that resides in it. Principally, I think any functionality would be paled of the StepRepository other than just fetching resources, since it will not be included in the unit tests. For now it's not a big deal, since the functionality is so limited,  but I just thought that in future someone may be tempted to put more functionality in it. On the other hand, because there is so little functionality today, making a class for it would be a little awkward.

test/unit/clarity_ext/dilution/test_dilutions.py
withrocks: That's a valid concern. 

My idea is that the code in the private repo should generally do almost nothing, only format, reorder and so on - specifics for the configuration of the devices we have. So it should perhaps read more like configuration rather than code. Those specifics will not be likely to be reused in exactly that format (if they ever are, I think the feature should then be pushed down to clarity-ext). In this case, the code is actually partially a copy of the extension code, but the beauty is that the extension can change at will, while this test would not change, still validating that `DilutionScheme` actually is doing the correct thing.

The power in the split between the library and the particular extensions (configuration) built from it will only be clear if anyone ever uses the underlying library. Although I'm not sure anyone else will ever use it, I am sure that no-one will use it if it's not correctly broken up from the beginning (e.g. if it will contain a ton of code that's only interesting for us.)

The code that I did copy is the order of the values and the formatting. I would actually prefer that this test would not do that formatting, then it would not really be duplication. I kept it in there to make it clear that this actually tests what the old code tested - but I think it wasn't valid, so I'll remove the formatting to make this clear.

withrocks: ... so I think of this kind of duplication as "forking" rather than just blind duplication, since each part will live their separate lives from then on.

test/unit/clarity_ext/dilution/test_dilutions.py
withrocks: It would be cool to have tests for the StepRepository, I think, but I think it's important to keep it super simple, only a wrapper around the REST proxy. I might be misunderstanding though.

test/unit/clarity_ext/dilution/test_dilutions.py
withrocks: Yes, I agree. It's difficult to find that level, I think.

In this case, I feel it's doing the right thing. I think of it this way: Here we are only doing *one thing* that actually does something, i.e. setting up the object and then viewing it through the `dilutes` attribute. That I think should correspond to one test, since otherwise, one is doing that same action over and over again. I'm not so much concerned about the performance implications of that, but rather the maintainability.

I would though argue that it should have more asserts (but in one test), asserting each fact that should be true separately. Then, when the test fails in the future, it would be clear which part of the test failed. I think that would be excellent in a refactoring session.

withrocks:
You may merge this now if there are no more comments.

clarity_ext/domain/analyte.py
```diff
@@ -1,54 +1,54 @@
...
     """
-    Describes an Analyte in the Clarity LIMS system, including custom UDFs.
+    Describes an Analyte in the Clarity LIMS system.
 
-    Takes an analyte resource as input.
+    Expects certain mappings to UDFs in clarity. These are provided
+    in udf_map, so they can be overridden in different installations.
     """
 
-    def __init__(self, resource, container):
```
GitEdvard: Is kwargs restricted to udfs? In that case it should be indicated by the paramter name.

clarity_ext/domain/analyte.py
```diff
@@ -1,54 +1,54 @@
...
-    def well(self):
-        pos = PlatePosition.create(self.resource.location[1])
-        return Well(pos, self.container)
 
-    @property
-    def sample(self):
-        return self.resource.samples[0]
+class Sample(object):
+    def __init__(self, **kwargs):
+        self.id = kwargs.get("id", None)
```
GitEdvard: Isn't it better to restrict kwargs to udfs? The parameter "id" is something that is not changing between different Clarity setups, and the parameter kwargs is a bit convoluted.

clarity_ext/domain/analyte.py
withrocks: My idea was not that it should be restricted to UDFs, but rather that it should contain all non-positional arguments, which happen to include UDFs too. As I see it, the domain object shouldn't necessarily know about UDFs as a Clarity concept. Only the `create_from_rest_resource` needs to know about that, that's why it's that method that has the udf_map.

So this is just a kwargs dictionary and it's a custom to use the name kwargs for that (keyword arguments) in Python. The idea is that listing all the possible positional arguments in the constructor becomes cumbersome:

https://github.com/search?l=Python&q=kwargs&type=Repositories&utf8=%E2%9C%93

clarity_ext/domain/analyte.py
```diff
@@ -1,54 +1,54 @@
...
-    def well(self):
-        pos = PlatePosition.create(self.resource.location[1])
-        return Well(pos, self.container)
 
-    @property
-    def sample(self):
-        return self.resource.samples[0]
+class Sample(object):
+    def __init__(self, sample_id):
+        self.id = sample_id
```
withrocks: Good point. I've changed it, since it always makes sense to send in an ID when creating a sample.

It's difficult to say when one should take parameters as keyword arguments and when not. In this case, I think it's valid that it should be positional.

GitEdvard:
Ok

PR #24, opened 2016-08-19
Run test coverage when running the tests

withrocks:
- Running the validate scripts now creates coverage files in ./cover
  - Added a test for the lazyprop decorator
  - Fixed some comments
  - Add a script for cleaning up the syntax of all files changed in a
    branch.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ from setuptools import find_packages, setup
 
 # Add that to the dev reqs file for now, since it's not needed when running
 dependencies = ['click', 'genologics', 'requests-cache', 'pyyaml', 'nose', 'PyPDF2',
-                'lxml']
+                'lxml', 'coverage', 'pep8radius']
 
 setup(
     name='clarity-ext',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -5,7 +5,7 @@
 
 # Add that to the dev reqs file for now, since it's not needed when running
 dependencies = ['click', 'genologics', 'requests-cache', 'pyyaml', 'nose', 'PyPDF2',
-                'lxml']
+                'lxml', 'coverage', 'pep8radius']
```
GitEdvard: Ok, so if I use u new package, I should add it here?

GitEdvard:
Ok, I merge this!

setup.py
withrocks: Yes, for now, they're listed in setup.py. We'll probably want to use a requirements file in the future, so people don't have to install dev requirements in production, e.g nose, coverage and pep8radius, but I'll clean that up later.

(The other repo (that uses this) is however setup with a conda metadata file on the prod servers.)

PR #27, opened 2016-09-07
Lims 655 bugfix

GitEdvard:
Fix so that all integration tests work:
* Adapt _repr_ of well to handle if the container is not set. 
* Do not cast exception for ResultFile in update_objects in step_repository. Replace with TODO
Prevent an infinite loop when comparing equality for objects containing mutual references, e.g. Analyte --> well and Well --> artifact.

### DIFF
--- a/clarity_ext/domain/common.py
+++ b/clarity_ext/domain/common.py
@@ -1,10 +1,40 @@
+from __builtin__ import isinstance
+
+
 class DomainObjectMixin(object):
 
     def __eq__(self, other):
         if isinstance(other, self.__class__):
-            return self.__dict__ == other.__dict__
+            return self._eq_rec(self, other)
         else:
             return False
+
+    def _eq_rec(self, a, b, cache=[]):
+        """
+        Replaces the == operator because of circulating references (e.g. analyte <-> well)
+        Adapted solution taken from
+        http://stackoverflow.com/questions/31415844/using-the-operator-on-circularly-defined-dictionaries
+        """
+        cache = cache + [a, b]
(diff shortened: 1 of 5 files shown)


clarity_ext/domain/common.py
```diff
@@ -1,10 +1,40 @@
...
 class DomainObjectMixin(object):
 
     def __eq__(self, other):
         if isinstance(other, self.__class__):
-            return self.__dict__ == other.__dict__
+            return self._eq_rec(self, other)
         else:
             return False
 
+    def _eq_rec(self, a, b, cache=[]):
```
withrocks: Interesting. I guess that another way to solve this is to not use __dict__ comparison, but instead write an equality operator for each domain object. I think that might be cleaner anyways. But I think going for the generic solution is fine for now at least.

clarity_ext/repository/step_repository.py
```diff
@@ -131,6 +131,9 @@ def update_objects(self, objects):
             if type(obj) == Analyte:
                 original_analyte_from_rest = ApiArtifact(self.session.api, id=obj.id)
                 update_queue.append(obj.updated_rest_resource(original_analyte_from_rest, self.udf_map))
+            elif type(obj) == ResultFile:
```
withrocks: I generally prefer to throw NotImplemented when something isn't implemented, and instead change the tests/code that's using this to not do that until it's ready (fail-fast). Is there some reason that's not possible in this case?

withrocks:
Merging. The TODO will be fixed soon (TM) by Edvard.

PR #29, opened 2016-09-09
Ensure ResultFile UDFs are updated on commit

withrocks:
Fixes LIMS-661

### DIFF
--- a/clarity_ext/domain/result_file.py
+++ b/clarity_ext/domain/result_file.py
@@ -19,6 +19,9 @@ class ResultFile(Artifact):
             value = self.units.convert(value, from_unit, to_unit)
         self.api_resource.udf[name] = value
 
+    def get_udf(self, name):
+        return self.api_resource.udf[name]
+
     @staticmethod
     def create_from_rest_resource(resource, container_repo):
         """
@@ -27,7 +30,6 @@ class ResultFile(Artifact):
         """
         ret = ResultFile(api_resource=resource,
                          units=UnitConversion(), id=resource.id)
-        ret._resource = resource
 
         try:
             container_resource = resource.location[0]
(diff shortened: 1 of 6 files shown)


clarity_ext/domain/result_file.py
```diff
@@ -19,6 +19,9 @@ def set_udf(self, name, value, from_unit=None, to_unit=None):
             value = self.units.convert(value, from_unit, to_unit)
         self.api_resource.udf[name] = value
 
+    def get_udf(self, name):
+        return self.api_resource.udf[name]
+
```
GitEdvard: As far as I can see, this api resource is not in synch with the domain object (result_file). The api resource has the state as  fetched first time from db.

GitEdvard: Ok, I looked another time, this is in order.

clarity_ext/repository/step_repository.py
```diff
@@ -121,23 +121,13 @@ def _wrap_artifacts(self, artifacts):
...
-                update_queue.append(obj.updated_rest_resource(original_analyte_from_rest, self.udf_map))
-            elif type(obj) == ResultFile:
-                # TODO: Implement
-                pass
-            else:
-                raise Exception("Unknown type {}".format(obj.type))
-
-        self.session.api.put_batch(update_queue)
+        api_resources = [obj.api_resource for obj in artifacts]
+        self.session.api.put_batch(api_resources)
```
GitEdvard: But here, the api-resource is not in synch. There will just be a update with original values.

withrocks: It is if the domain object has ensured that the underlying resource is updated, which I do in the case of `ResultFile`s and I think all domain objects should do (take care of themselves). This can be seen by running `clarity-ext/test/integration/test_context.py:test_can_update_udfs`.

Is it not being updated in the case of analytes?

GitEdvard:
Ok

clarity_ext/repository/step_repository.py
withrocks: Note: We synched on this, and we're going to merge this solution with Edvard's in 
https://github.com/Molmed/clarity-ext/pull/28

PR #30, opened 2016-09-14
Two solutions for updating db

GitEdvard:
The latter solution works if we have a consensus of:
* Updates of udf is always done in the scripts, with the function set_udf 

wont work if:
* using ordinary assignment of instance variables, then calling commit().

I have a real hard time accepting this last solution for the fact that it's not the state of the domain object that is updated. Kind of chilling my nervous system.

### DIFF
--- a/clarity_ext/domain/analyte.py
+++ b/clarity_ext/domain/analyte.py
@@ -12,12 +12,11 @@ class Analyte(Artifact):
     in udf_map, so they can be overridden in different installations.
     """
 
-    def __init__(self, api_resource, name=None, well=None, sample=None,
-                 artifact_specific_udf_map=None, id=None, **kwargs):
+    def __init__(self, api_resource, name=None, well=None, sample=None, id=None, **kwargs):
         """
         Creates an analyte
         """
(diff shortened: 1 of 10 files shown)


clarity_ext/domain/analyte.py
```diff
@@ -68,18 +69,17 @@ def updated_rest_resource(self, original_rest_resource, udf_map, updated_fields)
...
         values_by_udf_names = {analyte_udf_map[key]: self.__dict__[key]
                                for key in analyte_udf_map if key in updated_fields}
-        assigner = AssignLogger(self)
         # Retrieve fields that are updated, only these field should be included in the rest update
         for key in values_by_udf_names:
             if _updated_rest_resource.udf.get(key, None):
                 original_type = type(_updated_rest_resource.udf[key])
                 value = get_and_apply(values_by_udf_names, key, None, original_type)
-                _updated_rest_resource.udf[key] = assigner.log_assign(key, value)
+                _updated_rest_resource.udf[key] = self.assigner.log_assign(key, value)
```
withrocks: Minor: It's a little bit confusing for me to read `log` when it's something that has to do with business-logic rather than being just a straightforward logger.

PR #31, opened 2016-09-15
Make unit tests for file services

GitEdvard:
Make unit tests for
* Driver file service
* Response file service

First, extracting os functions from general file service in order to avoid them in unit tests.

This PR implements LIMS662 and LIMS668


clarity_ext/extensions.py
```diff
@@ -147,21 +148,17 @@ def execute(self, module, mode, run_arguments_list=None, config=None, artifacts_
...
                 context = ExtensionContext.create(
                     run_arguments["pid"], cache=cache_artifacts)
-                file_svc = None
+                instance = extension(context)
+                os_service = OSService()
                 if issubclass(extension, DriverFileExtension):
-                    instance = extension(context)
-                    driver_file_svc = DriverFileService(instance, self.logger)
-                    file_svc = GeneralFileService(driver_file_svc, ".")
+                    file_svc = DriverFileService.create_file_service(instance, self.logger, os_service)
```
withrocks: Just a minor point: I like when these static factory methods on services know how the instance should be created in production code, so in this case I would consider having the factory method also instantiate the OSService. Then the user only needs to use the factory method to create a production ready instance, which is nice to have in integration tests too.

PR #33, opened 2016-09-19
Integration test fixes

GitEdvard:
Fix #1:
Fix shared file names for integration tests. Function shared_file in extension is to be used in production and be able to generate a file name based on current date and user initials, whereas integration tests use a shared file name already created and should therefore be hardcoded. 

Fix #2:
Possibility for preparation of integration tests. Source volume in dilutions are updated in a relative basis (current volume - dilute volume) and are therefor changing for each integration test run. Here is a class for running preparation code before the integration test is run. In this case, the volumes are updated to predefined values.

### DIFF
--- a/test/integration/integration_test_service.py
+++ b/test/integration/integration_test_service.py
@@ -10,16 +10,15 @@ class IntegrationTest:
             self.run_argument_dict.update(run_argument_dict)
 
         if update_matrix:
-            self.preparer = IntegrationTestPrepare(update_matrix, self.pid)
+            self.preparer = IntegrationTestPrepare(update_matrix)
 
     def pid(self):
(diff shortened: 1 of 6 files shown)


test/integration/integration_test_service.py
```diff
@@ -0,0 +1,49 @@
...
+        for update_row in self.update_matrix:
+            art_id = update_row[0]
+            if art_id not in artifact_dict:
+                raise ArtifactsNotFound("Given lims-id is not matching artifacts in step ({})".format(art_id))
+
+
+class ArtifactsNotFound(Exception):
+    pass
+
+
```
withrocks: These extra lines would be automatically removed by running `pep8radius`.

PR #40, opened 2016-09-23
Lims 708 source only dilution scheme

GitEdvard:
A dilution class is created, source only dilution scheme,  adapted for a dilution with no output analytes (qPCR). In this step there are only sample measurements as outputs (result_file). These changes have been made:
* Add class for user
* Refactor the Dilute class. Make it deal with units for input and output analytes separately.
* Refactor RobotDeckPositioner. Make it use two instances of the new class EndpointPositioner, which handles all plates for either source or destination.

### DIFF
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -126,6 +126,17 @@ class ExtensionContext(object):
     def output_result_files(self):
         return self.artifact_service.all_output_files()
 
+    @property
+    def user_initials(self):
+        return self.current_user.initials
+
+    @property
+    def pid(self):
+        return self.session.current_step_id
+
+    def today(self, format):
+        return datetime.date.today().strftime(format)
+
     def update(self, obj):
         """Add an object that has a commit method to the list of objects to update"""
         self._update_queue.append(obj)
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -132,15 +132,8 @@ class ExtensionContext(object):
         return self.artifact_service.all_output_files()
 
     @property
-    def user_initials(self):
-        return self.current_user.initials
-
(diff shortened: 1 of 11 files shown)


clarity_ext/context.py
```diff
@@ -124,6 +131,17 @@ def output_result_file_by_id(self, file_id):
     def output_result_files(self):
         return self.artifact_service.all_output_files()
 
+    @property
+    def user_initials(self):
```
withrocks: I think we should start thinking about limiting the properties we add to the context, as it will otherwise start getting rather crowded. So in this case, perhaps it would suffice that the extension writer could write `context.current_user.initials` instead of having a property for that.

clarity_ext/context.py
```diff
@@ -124,6 +131,17 @@ def output_result_file_by_id(self, file_id):
...
+        return self.current_user.initials
+
+    @property
+    def pid(self):
+        return self.session.current_step_id
+
+    def today(self, format):
```
withrocks: I wonder the same here as above. I don't think we need to hide too much of core Python stuff for the user. The main idea with the `context` is I believe to hide boilerplate related to Clarity. I think we will otherwise go down the rabbit hole of proxying everything Python has to offer.

GitEdvard: Yes, I was considering that fact when writing. I had in mind to get as simple as possible in the scripts, but I think it's maybe better with the suggestion above. I'll change it before the merge.

GitEdvard:
Not a big issue, but I kept the pid method. I use to have some problems finding it in session.

withrocks:
Yes, I think that one makes a lot of sense actually.

PR #41, opened 2016-09-27
Lims 686 multiple rows for high volume transfers

GitEdvard:
In driver file generation, split up rows containing more than 50 ul of pipetting volume, either sample volume or buffer volume. 

In addition, scale up sample volumes less than 2 ul (min pipetting volume). This means that the entered target volume from user is also scaled up. 

Small refactoring: Sorting of the transfers are broken out from do_posiitoning, and called explicitly in __init__. Sorting also shows splitted up transfers in an intuitive order, higher volumes first.

### DIFF
--- a/clarity_ext/dilution.py
+++ b/clarity_ext/dilution.py
@@ -1,7 +1,10 @@
 from clarity_ext.domain.validation import ValidationException, ValidationType
+from clarity_ext.utils import lazyprop
+import copy
 
 DILUTION_WASTE_VOLUME = 1
 ROBOT_MIN_VOLUME = 2
+PIPETTING_MAX_VOLUME = 50
 
 
 class TransferEndpoint(object):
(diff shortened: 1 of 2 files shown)


clarity_ext/dilution.py
```diff
@@ -242,11 +250,84 @@ def calculate_transfer_volumes(self):
...
+                self._get_volume(transfer.sample_volume), self._get_volume(transfer.buffer_volume))
+            (n, residual) = divmod(calculation_volume, PIPETTING_MAX_VOLUME)
+            if residual > 0:
+                duplication_number = int(n + 1)
+            else:
+                duplication_number = int(n)
+
+            # Copy transfer x times and add to array
+            # Update sample volume and buffer volume if needed
+            splitted_transfers = self._split_up_single_transfer(
```
withrocks: Note: I believe it's just "split", not splitted.

clarity_ext/dilution.py
```diff
@@ -242,11 +250,84 @@ def calculate_transfer_volumes(self):
...
+                transfer, duplication_number)
+            self.transfers += splitted_transfers
+
+    @staticmethod
+    def _split_up_single_transfer(transfer, number_rows):
+        transfers = []
+        for i in range(0, number_rows):
```
withrocks: Not explicitly necessary here, but I suggest always using `xrange` rather than `range`. The latter materializes a list unnecessarily.

clarity_ext/dilution.py
```diff
@@ -242,11 +250,84 @@ def calculate_transfer_volumes(self):
...
+            if i == number_rows - 1:
```
withrocks: It's a little bit hard to follow the logic here, in my mind. Partially since there is no comment on what the method is supposed to do, but I think the readability could be increased. I don't think it's necessary now though.

clarity_ext/dilution.py
```diff
@@ -242,11 +250,84 @@ def calculate_transfer_volumes(self):
...
+        return transfers
+
+    def sort_transfers(self):
+        def pipetting_volume(transfer):
+            return self._get_volume(transfer.buffer_volume) + self._get_volume(transfer.sample_volume)
+
+        def max_added_pip_volume():
+            volumes = map(lambda t: (self._get_volume(t.buffer_volume),
+                                     self._get_volume(t.sample_volume)), self.transfers)
+            return max(map(lambda v: v[0] + v[1], volumes))
```
withrocks: For readability, I prefer to use unpacking rather than indexing in statements like these, since it will make it clear to the reader what these are, e.g.:

`lambda buffer_vol, sample_vol: buffer_vol + sample_vol, volumes`

PR #43, opened 2016-09-28
Add support for logging

withrocks:
- Ensures that all calls to the core logging framework end up
  in /opt/clarity-ext/logs on a production server.
- Allows extension writers to write to log files on the step easily

### DIFF
--- a/clarity_ext/service/step_logger_service.py
+++ b/clarity_ext/service/step_logger_service.py
@@ -1,0 +1,60 @@
+import logging
+import time
+from clarity_ext.service.file_service import SharedFileNotFound
+
+
+class StepLoggerService:
+    """
+    The logger service provides logging for files attached in the step.
+
+    Extension writers are exposed to this class via the `ExtensionContext`, which supplies direct
+    access to both logging artifacts, without them having to do anything but ensuring that
+    a file is defined on the step
+    """
+    def __init__(self, step_logger_name, file_service, raise_if_not_found=False, append=True):
+        self.core_logger = logging.getLogger(__name__)
+        self.step_logger_name = step_logger_name
+        self.file_service = file_service
+
+        # Use Windows line endings for now, since most clients are currently Windows.
+        # TODO: This should be configurable.
+        self.NEW_LINE = "\r\n"
+        try:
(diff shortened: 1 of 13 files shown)


clarity_ext/service/step_logger_service.py
```diff
@@ -0,0 +1,60 @@
+import logging
+import time
+from clarity_ext.service.file_service import SharedFileNotFound
+
+
+class StepLoggerService:
+    """
+    The logger service provides logging for files attached in the step.
+
```
GitEdvard: Shouldn't it be 'logging to files' ... ?

clarity_ext/service/step_logger_service.py
```diff
@@ -0,0 +1,60 @@
...
+    Extension writers are exposed to this class via the `ExtensionContext`, which supplies direct
+    access to both logging artifacts, without them having to do anything but ensuring that
```
GitEdvard: the word 'both' seem redundant in 'access to both logging', or am I missing something?

withrocks: Yes, this makes no sense - probably remains of an old sentence. I'll rewrite the whole doc comment.

clarity_ext/service/step_logger_service.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    def __init__(self, step_logger_name, file_service, raise_if_not_found=False, append=True):
+        self.core_logger = logging.getLogger(__name__)
+        self.step_logger_name = step_logger_name
+        self.file_service = file_service
+
+        # Use Windows line endings for now, since most clients are currently Windows.
+        # TODO: This should be configurable.
+        self.NEW_LINE = "\r\n"
+        try:
```
GitEdvard: FYI: There is a new line method in GeneralFileExtension as well

withrocks: Yes, the idea with that method was that it specifies what type of newlines there should be used in driverfiles, specified by extension writers. I don't see any clear reason for reusing it though, since this is to be configured by administrators.

GitEdvard:
ok

PR #44, opened 2016-10-05
Lims 717 updated qpcr

GitEdvard:
Prepare clarity ext dilutions for a step configured with input analytes and output result files. (The UI name for result file is sample measurement, which is a more appropriate name for it. ) The input analyte and output result file are forming a pair and should be treated in the same way as an analyte pair. I have created an common base class for them, Aliquot, which is now used in DilutionScheme class. 

Also, make some fixes regarding Well instantiation. Add some tests for create_from_rest_resource (result_file and analyte), and updated_rest_resource

### DIFF
--- a/clarity_ext/dilution.py
+++ b/clarity_ext/dilution.py
@@ -1,6 +1,6 @@
 from clarity_ext.domain.validation import ValidationException, ValidationType
-from clarity_ext.utils import lazyprop
 import copy
+from clarity_ext.utils import get_and_apply
 
 DILUTION_WASTE_VOLUME = 1
 ROBOT_MIN_VOLUME = 2
@@ -13,17 +13,18 @@ class TransferEndpoint(object):
     """
     # TODO: Handle tube racks
 
-    def __init__(self, analyte):
-        self.sample_name = analyte.name
-        self.well = analyte.well
-        self.container = analyte.container
-        self.concentration = analyte.concentration
-        self.volume = analyte.volume
-        self.requested_concentration = analyte.target_concentration
-        self.requested_volume = analyte.target_volume
+    def __init__(self, aliquot):
+        self.sample_name = aliquot.name
+        self.well = aliquot.well
+        self.container = aliquot.container
+        self.concentration = aliquot.concentration
+        self.volume = aliquot.volume
+        self.requested_concentration = get_and_apply(
+            aliquot.__dict__, "target_concentration", None, float)
+        self.requested_volume = get_and_apply(
+            aliquot.__dict__, "target_volume", None, float)
         self.well_index = None
         self.plate_pos = None
-
 
 class SingleTransfer(object):
     # Enclose sample data, user input and derived variables for a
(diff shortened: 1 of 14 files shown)


clarity_ext/dilution.py
```diff
@@ -13,18 +13,19 @@ class TransferEndpoint(object):
...
-        self.requested_concentration = analyte.target_concentration
-        self.requested_volume = analyte.target_volume
+    def __init__(self, aliquot):
+        self.sample_name = aliquot.name
+        self.well = aliquot.well
+        self.container = aliquot.container
+        self.concentration = aliquot.concentration
+        self.volume = aliquot.volume
+        self.requested_concentration = get_and_apply(
+            aliquot.__dict__, "target_concentration", None, float)
```
withrocks: The code would be simpler if the aliquot would just have `target_concentration` of type `float`.

clarity_ext/domain/result_file.py
```diff
@@ -1,35 +1,35 @@
-from clarity_ext.unit_conversion import UnitConversion
-from clarity_ext.domain.artifact import Artifact
+from clarity_ext.domain.aliquot import Aliquot, Sample
 
 
-class ResultFile(Artifact):
+class ResultFile(Aliquot):
```
withrocks: Why does it now inherit form Aliquot? It seems unnecessarily specific to me, since it can be any file - perhaps not in the step you're working with, but generally. See definition here:

"ResultFile: The process outputs a file, such as a result file or a file produced by an instrument."
https://genologics.zendesk.com/hc/en-us/articles/213973183-Creating-Configuring-Processes

The reason for why it is inheriting from `Artifact` was that it needs to know if it's an input artifact and have UDFs associated to it.

clarity_ext/repository/step_repository.py
```diff
@@ -171,5 +174,9 @@ def current_user(self):
         "target_concentration": "Target Concentration",
         "target_volume": "Target Volume",
         "volume": "Current sample volume (ul)"
+    },
+    "ResultFile": {
```
withrocks: Here, I again think this is not really something that has to do with ResultFiles in general, but this ResultFile you're currently working with, while the other map had to do with Analytes.

I think you should probably think about having an object that inherits from ResultFile which has these properties.

clarity_ext/domain/result_file.py
GitEdvard: A result_file in clarity, (not a shared result file) is a rather poor name for it. When it's labelled as 'PerInput' it's about a sample measurement, that has a result file attached to it. Even if Analyte and Result file may have a rather different set of UDFs attached to it, they have sample placement in common. Regarding concentration and volume, they are a very basic measures that also is common to the two classes. 

I'm not convinced that it would be the right thing to cancel the Aliquot class. And certainly not now. 

UDF may be attached to every entity in Clarity, not only artifacts.

clarity_ext/repository/step_repository.py
GitEdvard: It's not even certain that volume and concentration are relevant for Analytes in all steps, but in some steps it's needed. According to teh reasoning above, there should be separate classes, depending on which step the analytes/result files are in. This is a relevant question however, the fact that Analyte and ResultFile will grow very big as they will cover all different kind of steps. The different UDFs only have meaning for a certain step.

withrocks:
Merged with techdebt ticket: https://snpseq.atlassian.net/browse/LIMS-758

PR #45, opened 2016-10-07
Fix for some broken tests

withrocks:
An earlier refactoring, which added the SharedResultFile
domain object, seems to have broken three tests.
This fixes that.


withrocks:
This breaks a test in `clarity-snpseq`. Will reopen another one that fixes both.

PR #46, opened 2016-10-10
Fix for some broken tests

withrocks:
An earlier commit, seems to have broken three tests.

- In one case an assert has been removed. It ran successfully in the clarity-ext tests but not in the clarity-snpseq tests. It is not required.

### DIFF
--- a/test/integration/domain/test_result_file.py
+++ b/test/integration/domain/test_result_file.py
@@ -1,6 +1,5 @@
-import requests_cache
 import unittest
-from clarity_ext.domain import Container, ResultFile
+from clarity_ext.domain import Container, ResultFile, SharedResultFile
 from clarity_ext.context import ExtensionContext
 import os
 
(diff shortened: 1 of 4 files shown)


test/integration/domain/test_result_file.py
```diff
@@ -54,7 +53,7 @@ def test_can_fetch_single_output_file(self):
         context = ExtensionContext.create("24-3144")
         result = context.output_result_file_by_id("92-5244")
         self.assertIsNotNone(result)
-        self.assertIsInstance(result, ResultFile)
+        self.assertIsInstance(result, SharedResultFile)
```
johanherman: Should this really be a `SharedResultFile`? 

Because if I trace the execution path `context.output_result_file_by_id() -> artifact_service.output_file_by_id() -> all_output_files()` I see an `assert` for a `ResultFile` on line 91 (https://github.com/Molmed/clarity-ext/blob/432d4304d6c1040b6b0e0a8f76e09d4c70895829/clarity_ext/service/artifact_service.py#L91), as well as an comparison of the type with `Artifact.OUTPUT_TYPE_RESULT_FILE`. 

There's an `all_shared_result_files()` with a comparison with the `Artifact.OUTPUT_TYPE_SHARED_RESULT_FILE` type instead, but this also has an `assert` for `ResultFile`. 

So I don't really get how this fits together.

withrocks: Yes, I agree that this needs to be looked into.

withrocks:
Closed with tech-debt ticket.

PR #48, opened 2016-10-12
Support generation of samplesheet

withrocks:
This pull request provides support for generating samplesheets. It also contains various related fixes and refactorings.

### DIFF
--- a/clarity_ext/domain/aliquot.py
+++ b/clarity_ext/domain/aliquot.py
@@ -6,12 +6,12 @@ from clarity_ext.domain.container import ContainerPosition
 
 class Aliquot(Artifact):
 
-    def __init__(self, api_resource, is_input, id=None, sample=None, name=None, well=None,
+    def __init__(self, api_resource, is_input, id=None, samples=None, name=None, well=None,
                  artifact_specific_udf_map=None, **kwargs):
         super(Aliquot, self).__init__(
             api_resource=api_resource, id=id, name=name,
             artifact_specific_udf_map=artifact_specific_udf_map)
-        self.sample = sample
+        self.samples = samples
         self.well = well
         self.is_input = is_input
         if well:
--- a/clarity_ext/domain/aliquot.py
+++ b/clarity_ext/domain/aliquot.py
@@ -46,13 +46,22 @@ class Aliquot(Artifact):
 
 class Sample(DomainObjectMixin):
 
-    def __init__(self, sample_id):
+    def __init__(self, sample_id, name, project):
         self.id = sample_id
+        self.name = name
+        self.project = project
 
     def __repr__(self):
         return "<Sample id={}>".format(self.id)
 
     @staticmethod
     def create_from_rest_resource(resource):
-        sample = Sample(resource.id)
+        project = Project(resource.project.name) if resource.project else None
+        sample = Sample(resource.id, resource.name, project)
         return sample
+
+
+class Project(DomainObjectMixin):
+    def __init__(self, name):
+        self.name = name
+
--- a/clarity_ext/domain/analyte.py
+++ b/clarity_ext/domain/analyte.py
@@ -51,6 +51,7 @@ class Analyte(Aliquot):
                           samples=samples, name=resource.name,
                           well=well, artifact_specific_udf_map=analyte_udf_map, **kwargs)
         analyte.api_resource = resource
+        analyte.reagent_labels = resource.reagent_labels
 
         return analyte
 
--- a/clarity_ext/domain/analyte.py
+++ b/clarity_ext/domain/analyte.py
@@ -1,5 +1,6 @@
 from clarity_ext.utils import get_and_apply
 from clarity_ext.domain.aliquot import Aliquot, Sample
+from clarity_ext import utils
 
 
 class Analyte(Aliquot):
@@ -10,13 +11,13 @@ class Analyte(Aliquot):
     in udf_map, so they can be overridden in different installations.
     """
 
-    def __init__(self, api_resource, is_input, id=None, sample=None, name=None, well=None,
+    def __init__(self, api_resource, is_input, id=None, samples=None, name=None, well=None,
                  artifact_specific_udf_map=None, **kwargs):
         """
         Creates an analyte
         """
         super(self.__class__, self).__init__(api_resource, is_input=is_input, id=id,
-                                             sample=sample, name=name, well=well,
+                                             samples=samples, name=name, well=well,
                                              artifact_specific_udf_map=artifact_specific_udf_map, **kwargs)
         self.target_concentration = get_and_apply(
             kwargs, "target_concentration", None, float)
@@ -45,9 +46,9 @@ class Analyte(Aliquot):
 
         # TODO: sample should be put in a lazy property, and all samples in a step should be
         # loaded in one batch
-        sample = Sample.create_from_rest_resource(resource.samples[0])
+        samples = [Sample.create_from_rest_resource(sample) for sample in resource.samples]
         analyte = Analyte(api_resource=resource, is_input=is_input, id=resource.id,
-                          sample=sample, name=resource.name,
+                          samples=samples, name=resource.name,
                           well=well, artifact_specific_udf_map=analyte_udf_map, **kwargs)
         analyte.api_resource = resource
 
@@ -66,4 +67,13 @@ class Analyte(Aliquot):
         if 'name' in updated_fields:
             _updated_rest_resource.name = self.assigner.register_assign('name', self.name)
         return _updated_rest_resource, self.assigner.consume()
+
+    @property
+    def sample(self):
+        """
+        Returns a single sample for convenience. Throws an error if there isn't exactly one sample.
+
+        NOTE: There can be more than one sample on an Analyte. That's the case with pools.
+        """
+        return utils.single(self.samples)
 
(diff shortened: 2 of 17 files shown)


clarity_ext/service/artifact_service.py
```diff
@@ -106,3 +118,45 @@ def all_shared_result_files(self):
...
+            ...
+
+        This method will fetch all of the input artifacts based on all of your output artifacts in one call
+        and index them by their respective process id.
+        """
+
+        # We will need the input artifacts from the previous step
+        parent_processes = set([artifact.input.parent_process for artifact in self.all_output_artifacts()])
+
+        for process in parent_processes:
```
johanherman: Perhaps you could explain for me what these processes are, how they and their trees are created. E.g. wondering if it would be possible that this would end up in an infinite loop, depending on how the tree is generated. Although I don't think you're fetching all parents in the framework, but only those directly above? And then you leave the rest up to the script? I haven't followed these function calls all the way though, so I might be wrong.

clarity_ext/domain/analyte.py
```diff
@@ -67,3 +68,12 @@ def updated_rest_resource(self, original_rest_resource, updated_fields):
...
         return _updated_rest_resource, self.assigner.consume()
 
+    @property
+    def sample(self):
+        """
+        Returns a single sample for convenience. Throws an error if there isn't exactly one sample.
+
+        NOTE: There can be more than one sample on an Analyte. That's the case with pools.
+        """
+        return utils.single(self.samples)
```
withrocks: The business case being solved: We need to support "pooled" analytes. If you want details, you can check out [this](http://zarlab.cs.ucla.edu/sequencing-with-dna-pools/) article, or ask the lab. Before this commit, this use case hadn't turned up yet, so it supported only one sample.

Rationale behind providing both `samples` and `sample`: A design decision behind the library is that all domain objects should be as easy to use as possible. Where to drag that line can be difficult, but the idea here is that the domain expert (maintaining the extension) can write either `analyte.samples[0]` or `analyte.sample`, depending on what they are expecting. If they do only expect there to be one sample, they don't need to write the guards themselves. The underlying code should take care of throwing an exception if required. (This has the additional benefit that I didn't need to refactor all usage of `sample` already in the codebase when introducing the change.)

It could be argued that the code would allow for there to be no sample at all (returning `None` when zero), but I don't believe that will be needed. It can easily be changed afterwards if required.

clarity_ext/service/artifact_service.py
```diff
@@ -88,7 +88,6 @@ def all_output_files(self):
         files = (outp for outp in outputs
                  if outp.output_type == Artifact.OUTPUT_TYPE_RESULT_FILE)
         ret = list(utils.unique(files, lambda f: f.id))
-        assert len(ret) == 0 or isinstance(ret[0], ResultFile)
```
withrocks: Yes, I think that needs to be addressed too, but decided to limit the effect of this commit.

This assert was failing because of an earlier commit, which needs to be cleaned up. I did only remove the assert that was causing tests to fail, and I added a ticket on the rest (LIMS-769). I think it would make sense to investigate that when working on the ticket.

clarity_ext/service/artifact_service.py
withrocks: So a `Process` is one step in a workflow protocol. There is some confusion about these sometimes being called steps and sometimes processes, but the API endpoint is `/processes`.

Yes, I only fetch for the parent process in this case, but future calls might try to fetch the entire tree. I think that you might be right that that would cause circular fetches, but I don't think it's a concern in this commit.

clarity_ext/extensions.py
```diff
@@ -351,6 +351,33 @@ def content(self):
...
+    """
+    __metaclass__ = ABCMeta
+
+    NONE = "<none>"
+
+    def header(self, name):
+        return self.line("[{}]".format(name))
+
+    def udf_prefixed(self, name, mapping=None):
+        return self.udf(name, mapping, prefix=name)
```
withrocks: You're right, I'll remove this!

clarity_ext/domain/analyte.py
johanherman: Aha. I guess that makes sense. I didn't realize the domain expert is expected to also write `analyte.samples[x]` if appropriate. Thanks. :+1:

clarity_ext/extensions.py
```diff
@@ -351,6 +351,33 @@ def content(self):
...
+    def udf(self, name):
+        """Returns the UDF if available, othervise self.NONE. Provided for readability"""
+        udf_val = self.context.udfs.get(name, self.NONE)
+        return udf_val
+
+    @staticmethod
+    def line(*args):
+        """Generates one line of the sample sheet, CSV formatted"""
+        # TODO: The example shows commas in each line. Is that actually required?
+        arg_list = list(args) + [""] * (9 - len(args))
```
withrocks: You're right, this has to do with the number of fields, but I seem forgot about refactoring this to be more generic (I don't like magic numbers either.)

I'll follow this up with another commit which hopefully makes it clearer.

johanherman:
I likw how you've named your branch according to the JIRA issue btw.

withrocks:
I've added one extra commit which is required by the code in `clarity-snpseq`, it got temporarily lost in a stash/rebase/cherry-picking adventure yesterday.

test/integration/integration_config.py
```diff
@@ -0,0 +1,24 @@
+import os
+import yaml
+
+
+class IntegrationConfig:
+    def __init__(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        path = os.path.join(dir_path, "clarity_steps.yml")
```
withrocks: So we have a rather inconvenient situation that we must do some integration tests against the LIMS server. Up until now, we've been hardcoding values, but now one can access them from a configuration file, which describes the data that needs to be in place on the server. This makes it easier for us to change it later or for external parties to set up the LIMS in such a state that they can validate it.

I use it in a later commit, here: 5da62ce, test/integration/service/test_artifact_service.py

clarity_ext/domain/aliquot.py
```diff
@@ -46,13 +46,22 @@ def create_well_from_rest(resource, container_repo):
...
+        self.name = name
+        self.project = project
 
     def __repr__(self):
         return "<Sample id={}>".format(self.id)
 
     @staticmethod
     def create_from_rest_resource(resource):
-        sample = Sample(resource.id)
+        project = Project(resource.project.name) if resource.project else None
```
withrocks: So the design is (on a high level) this:

- Clarity provides an API, from which we get the API resources. These are simple data contract objects. We use the Scilifelab/Stockholm library (genologics) for accessing those
- We got domain classes that may be "thick" business objects, which can be complex and contain references to other objects without issues. These will be used by the end user configuring the system.
- We get `repository` classes which fetch data from the API and map them to domain classes. The mapping happens through factory methods like the one we're looking at now.
- We got `services` which fetch data from the repositories. In general, unit testing them by faking only the repositories should be enough to test most of the business logic.

clarity_ext/domain/container.py
```diff
@@ -81,26 +81,33 @@ class Container(DomainObjectMixin):
     CONTAINER_TYPE_96_WELLS_PLATE = 100
     CONTAINER_TYPE_STRIP_TUBE = 200
     CONTAINER_TYPE_TUBE = 300
+    CONTAINER_TYPE_PATTERNED_FLOW_CELL = 400
```
withrocks: Yes, these are as you say IDs. The `size` mentioned in the commit log refers to the size of the container in the number of wells it contains (rows and columns.)

clarity_ext/domain/container.py
```diff
@@ -81,26 +81,33 @@ class Container(DomainObjectMixin):
...
-            self.size = PlateSize(height=8, width=1)
-        elif self.container_type == self.CONTAINER_TYPE_TUBE:
-            self.size = PlateSize(height=1, width=1)
+        if size:
+            self.size = size
         else:
-            raise ValueError("Unknown plate type '{}'".format(self.container_type))
+            # TODO: Require the size to be sent in instead
+            if self.container_type == self.CONTAINER_TYPE_96_WELLS_PLATE:
+                self.size = PlateSize(height=8, width=12)
```
withrocks: Yes, I think this will be removed entirely in a later commit, as these actually come from the LIMS. However, I didn't want to increase the scope of the commit (the commit is only ensuring that these will only be set when a size is not provided.)

test/integration/integration_config.py
johanherman: Ah. I missed that you did a `config = IntegrationConfig()` in ` test/integration/__init__.py`. :+1:

clarity_ext/driverfile.py
```diff
@@ -136,8 +136,8 @@ def file_key(self, file_name):
     def artifact(self):
         artifacts = [shared_file for shared_file in self.extension.context.shared_files
                      if shared_file.name == self.shared_file_name]
-        assert len(artifacts) == 1, "shared file {} could not be found".format(
-            self.shared_file_name)
+        assert len(artifacts) == 1, "Shared file '{}' not exactly single".format(
```
withrocks: So this particular change is only fixing the error message. But I believe my thinking at the time was that I didn't know if there would ever be more than one item in the list. The logic I had come up with needed the first item in the list, and in such cases I think it's prudent to throw an error or at least assert.

PR #49, opened 2016-10-17
Filter out blank controls in dilution scheme

GitEdvard:
Add boolean is_control to analyte, in order to filter out control samples.
Update map creation for mapping between AliquotTransfers and Aliquot pairs.

### DIFF
--- a/clarity_ext/dilution.py
+++ b/clarity_ext/dilution.py
@@ -230,9 +230,9 @@ class DilutionScheme(object):
                     transfer.buffer_volume *= scale_factor
                     transfer.scaled_up = True
             except (TypeError, ZeroDivisionError) as e:
-                transfer.sample_volume = 0
-                transfer.buffer_volume = 0
-                transfer.has_to_evaporate = False
+                transfer.sample_volume = None
+                transfer.buffer_volume = None
+                transfer.has_to_evaporate = None
 
     def split_up_high_volume_rows(self):
         """
--- a/clarity_ext/dilution.py
+++ b/clarity_ext/dilution.py
@@ -19,6 +19,9 @@ class TransferEndpoint(object):
         self.container = aliquot.container
         self.concentration = aliquot.concentration
         self.volume = aliquot.volume
+        self.is_control = False
+        if hasattr(aliquot, "is_control"):
+            self.is_control = aliquot.is_control
         self.requested_concentration = get_and_apply(
             aliquot.__dict__, "target_concentration", None, float)
         self.requested_volume = get_and_apply(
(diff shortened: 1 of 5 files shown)


clarity_ext/dilution.py
```diff
@@ -220,9 +230,9 @@ def calculate_transfer_volumes(self):
                     transfer.buffer_volume *= scale_factor
                     transfer.scaled_up = True
             except (TypeError, ZeroDivisionError) as e:
-                transfer.sample_volume = None
-                transfer.buffer_volume = None
-                transfer.has_to_evaporate = None
+                transfer.sample_volume = 0
```
johanherman: How does the logic around these volumes work? Are we supposed to silently ignore division by zero errors etc, or is/will it be handled elsewhere?

clarity_ext/domain/analyte.py
```diff
@@ -47,9 +48,14 @@ def create_from_rest_resource(resource, is_input, udf_map, container_repo):
         # TODO: sample should be put in a lazy property, and all samples in a step should be
         # loaded in one batch
         samples = [Sample.create_from_rest_resource(sample) for sample in resource.samples]
+        is_control = False
+        # TODO: This code principally belongs to the genologics layer, but 'control-type' does not exist there
+        if resource.root.find("control-type") is not None:
```
johanherman: Where *does* `control-type` come from? 

If I understand it correctly the `resource` object is from the Genologics layer that you're referencing in the comment, because if I tried to follow the function call path I seems to end up with a call to `genologics.entities.Process()` at https://github.com/Molmed/clarity-ext/blob/074944ac6848f53c79ec415c95a7ebccc41cf9bd/clarity_ext/clarity.py#L18

I just wanted to make sure that this `if` expression was sensible, but I don't know what properties `Process` returns, so I can't tell if `root` will always exist, nor can I see where `control-type` is set.

clarity_ext/dilution.py
johanherman: And then it is up to the script writer in `clarity-snpseq` to handle it in a sensible way?

clarity_ext/domain/analyte.py
johanherman: Ahh. OK. I don't have access to that URL at the moment, but if `root` is an XML property that always exists, and `find()` returns `None` when an element is not found (instead of with strings when it would be `-1` e.g.), then I guess I'm happy. :smile:

clarity_ext/dilution.py
GitEdvard: Ok, I remember now how my reasoning did go by this. I would like an error and warnings detection code to be run within the snpseq script. Some of these checks have to be run after the transfer volumes have been calculated. Exampels of checks: Is the volume too high? Do the sample have to be evaporated? Does all samples have a source concentration. The try-block in this method is to avoid the program to crash at the first faulty sample, so that instead a list av all faulty samples can be shown to the user.

The error and warnings checks is not implemented yet. The validation method in this module is eventually going to be removed.

johanherman: Ah. Gotcha. So it is fine that it is set to `None`, or might it cause problems elsewhere? I'm thinking that having the values set to `0` might be ambiguous, because perhaps sometimes it might make sense that it actually *should be* zero? On the other hand perhaps that should be left as an exercise for the script writer?

GitEdvard: I think either 0 or None will do for now. The reason I changed it was for debugging, the script crashed with assignment of None (to blank controls), and I wanted to see the ordinary output.

johanherman: All right. :+1:

PR #51, opened 2016-10-21
Add test for dilution with concentration unit nM

GitEdvard:
Referers to https://github.com/Molmed/clarity-snpseq/pull/39
Which is a bug fix regarding udf reference mixmatch


withrocks:
Cool. I've got pending changes which will hopefully make it easier to work with UDFs.

PR #53, opened 2016-10-28
Lims 725 warning popup

GitEdvard:
Break out validation code from DriverFileExtension to the new class
ValidationService. Move validation code in DilutionScheme to snpseq
scripts.

Errors and warnings generated from the validation are exported to a file with entry name Errors_and_warning. The script with the validation code are typically triggered from a button, to give the user the opportunity to change parameters until she is satisfied, clicking the button several times if needed. 

There is a check if the user tries to go to the next step with unresolved errors. For this, there is a udf 'Has errors2' associated with the artifact for the Shared result file 'Errors_and_warnings', which is updated when needed each time the button-script is run.

### DIFF
--- a/clarity_ext/dilution.py
+++ b/clarity_ext/dilution.py
@@ -204,11 +204,12 @@ class DilutionScheme(object):
     """Creates a dilution scheme, given input and output analytes."""
 
     def __init__(self, artifact_service, robot_name, scale_up_low_volumes=True,
-                 concentration_ref=None, include_blanks=False):
+                 concentration_ref=None, include_blanks=False, error_log_artifact=None):
         """
         Calculates all derived values needed in dilute driver file.
         """
         self.scale_up_low_volumes = scale_up_low_volumes
+        self.error_log_artifact = error_log_artifact
         pairs = artifact_service.all_aliquot_pairs()
 
         # TODO: Is it safe to just check for the container for the first output
--- a/clarity_ext/dilution.py
+++ b/clarity_ext/dilution.py
@@ -1,6 +1,6 @@
-from clarity_ext.domain.validation import ValidationException, ValidationType
 import copy
 from clarity_ext.utils import get_and_apply
+from itertools import groupby
 
 DILUTION_WASTE_VOLUME = 1
 ROBOT_MIN_VOLUME = 2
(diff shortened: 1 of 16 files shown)


clarity_ext/dilution.py
```diff
@@ -225,6 +228,18 @@ def __init__(self, artifact_service, robot_name, scale_up_low_volumes=True,
...
         self.sort_transfers()
+        self.grouped_transfers = list(self._grouped_transfers())
+
+    def _grouped_transfers(self):
+        """
+        Sometimes a sample transfer from source A to destination B is
+        split up on several rows, due to the max pipetting volume of 50 ul.
+        :return: An iterable group of transfers for a common destination well.
+        """
+        for key, transfer_group in groupby(
```
withrocks: One question: I assume that the data has already been sorted using the same key (since groupby generally requires it to be). Might break in marginal cases otherwise.

https://docs.python.org/2/library/itertools.html#itertools.groupby

clarity_ext/domain/shared_result_file.py
```diff
@@ -1,16 +1,30 @@
...
             api_resource=api_resource, id=id, name=name,
             artifact_specific_udf_map=artifact_specific_udf_map)
+        self.has_errors = bool(get_and_apply(kwargs, "has_errors", 0, int))
 
     @staticmethod
     def create_from_rest_resource(api_resource, udf_map=None):
-        shared_result_file_udf_map = udf_map.get("SharedResultFile")
+        shared_result_file_udf_map = udf_map.get("SharedResultFile", dict())
         name = api_resource.name
+        specific_udf_map = udf_map["SharedResultFile"]
```
withrocks: FYI: I have some ideas of removing the requirement of a UDF map and instead just use the underlying `udf_` properties which are now automatically mapped on the object. I'm confident that it will make a lot of things much simpler. It will be in the next milestone though.

clarity_ext/service/validation_service.py
```diff
@@ -0,0 +1,39 @@
...
+            for r in results:
+                msg_row = "{}".format(r)
+                self.step_logger_service.log(msg_row)
+                self._log_debug("{}".format(msg_row))
+
+        return self.has_errors, self.has_warnings
+
+    def _log_debug(self, msg):
+        if self.logger is not None:
+            self.logger.debug(msg)
```
withrocks: I'd just like to mention that another way would be to set the logger to an instance in the if not None, as in:

https://github.com/Molmed/clarity-ext/blob/b04e4f2a87b5cffda00ec9080cc3bb8f8cb61149/clarity_ext/driverfile.py#L38

PR #54, opened 2016-10-28
Add logic for supress printing out labels

GitEdvard:
At dilution steps (dna dilution + lib dilution), there is now a udf 'Print labels', that control if printing out labels or not, with default value 'no'. This udf should be shown during the validation period only.

In this pull request, make possible to get and set udfs for ExtensionContext. Previously, udfs could only be handled for Artifacts. 

Create the new base class Udf, and let both ExtensionContext and Artifact inherit from it. The code for udf handling is taken from the Artifact class.

### DIFF
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -8,9 +8,10 @@ from clarity_ext.repository import StepRepository
 from clarity_ext import utils
 from clarity_ext.driverfile import OSService
 from clarity_ext.service.validation_service import ERRORS_AND_WARNING_ENTRY_NAME
+from clarity_ext.domain.udf import Udf
 
 
-class ExtensionContext(object):
+class ExtensionContext(Udf):
     """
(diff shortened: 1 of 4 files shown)


clarity_ext/context.py
```diff
@@ -9,9 +7,10 @@
 from clarity_ext.repository import StepRepository
 from clarity_ext import utils
 from clarity_ext.driverfile import OSService
+from clarity_ext.domain.udf import Udf
 
 
-class ExtensionContext(object):
+class ExtensionContext(Udf):
```
withrocks: One might argue that this could be renamed for clarification (in a future commit), e.g. `SupportsUdfsMixin`.

withrocks: ... but as I mentioned in the other PR, I have some pending changes for UDFs, so I don't see any need for renaming this.

PR #55, opened 2016-11-02
Various changes to support CSV files

withrocks:
- CSV handling
- Other framework fixes (see individual commits)

### DIFF
--- a/clarity_ext/domain/container.py
+++ b/clarity_ext/domain/container.py
@@ -38,7 +38,11 @@ class Well(DomainObjectMixin):
 
 
 class ContainerPosition(namedtuple("ContainerPosition", ["row", "col"])):
-    """Defines the position of the plate, (zero based)"""
+    """
+    Defines the position of the plate, (zero based)
+
+    Default representation is `<row as letter>:<column as number>`, e.g. `A:1`
+    """
     def __repr__(self):
         return "{}:{}".format(self.row_letter, self.col)
 
@@ -170,6 +174,9 @@ class Container(DomainObjectMixin):
         return list(self.enumerate_wells(order))
 
     def set_well(self, well_pos, artifact_name=None, artifact_id=None, artifact=None):
+        # We should support any position that ContainerPosition can handle:
+        if not isinstance(well_pos, ContainerPosition):
+            well_pos = ContainerPosition.create(well_pos)
 
         if well_pos not in self.wells:
             raise KeyError(
@@ -180,8 +187,16 @@ class Container(DomainObjectMixin):
         # TODO: Accept only an artifact
         self.wells[well_pos].artifact = artifact
 
+    def __iter__(self):
+        for well in self.enumerate_wells(order=self.DOWN_FIRST):
+            yield well
+
+    def __setitem__(self, key, value):
+        self.set_well(key, artifact=value)
+
+    def __getitem__(self, item):
+        return self.wells[item]
 
     def __repr__(self):
-        return "{}".format(self.id)
-
+        return "Container(id={})".format(self.id)
 
--- a/clarity_ext/domain/container.py
+++ b/clarity_ext/domain/container.py
@@ -188,14 +188,18 @@ class Container(DomainObjectMixin):
         self.wells[well_pos].artifact = artifact
 
     def __iter__(self):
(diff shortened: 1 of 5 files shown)


clarity_ext/domain/container.py
```diff
@@ -180,8 +187,16 @@ def set_well(self, well_pos, artifact_name=None, artifact_id=None, artifact=None
         # TODO: Accept only an artifact
         self.wells[well_pos].artifact = artifact
 
+    def __iter__(self):
```
johandahlberg: Are you sure this works as expected? I'm not 100% what is the expected behavior  here but won't this get you a generator of a generator instead of looping over the initial generator?

Here is a small ipython experiment that might show my point:
```
In [1]: def inner_yield():
   ...:     yield 'a'
   ...:     

In [2]: def outer_yield():
   ...:     yield inner_yield()
   ...:     

In [3]: outer_yield()
Out[3]: <generator object outer_yield at 0x7f549103c320>

In [4]: list(outer_yield())
Out[4]: [<generator object inner_yield at 0x7f5491051af0>]

In [5]: list(list(outer_yield()))
Out[5]: [<generator object inner_yield at 0x7f5490f1fc80>]

In [6]: def outer_yield():
   ...:     return inner_yield()
   ...: 

In [7]: outer_yield()
Out[7]: <generator object inner_yield at 0x7f5491051aa0>

In [8]: list(outer_yield())
Out[8]: ['a']
```

So maybe I'm misunderstanding something here because I don't fully understand the domain, but I'd thought that you could just do `return self.enumerate_wells(order=self.DOWN_FIRST)`

withrocks:
Thanks for the comments. Regarding Travis, it has been the idea to turn it on after this left the prototype stage. I'll make sure to do it first thing in the next sprint!

FYI: In the private repo that implements the script for our organization (clarity-snpseq), we actually do have the concept of a limited set of "frozen" integration tests, where the responses from the server are cached. The reason for that feature was to be able to integration test without the original server. It's quite brittle though...

withrocks:
Regarding adding tests to handle the CSV logic, I agree. I needed this feature ASAP and apparently wasn't in unit test mode. I'll add a techdebt ticket on it.

clarity_ext/service/file_service.py
```diff
@@ -122,3 +127,40 @@ def cleanup(self):
 class SharedFileNotFound(Exception):
     pass
 
+
+class Csv:
+    """A simple wrapper for csv files"""
+    def __init__(self, file, delim=","):
```
withrocks: Yes, mine did too - I might have to turn that warning into an error to remember doing that. It's unfortunate here since this is actually expected to be a file-like object and this seems a good name for the instance. But, it is a built-in so I guess I'll just have to deal with that.

clarity_ext/service/file_service.py
```diff
@@ -122,3 +127,40 @@ def cleanup(self):
...
+        lines = list()
+        for line in file:
+            values = line.strip().split(delim)
+            csv_line = CsvLine(values, self)
+            lines.append(CsvLine(values, self))
+            if len(lines) == 1:
```
withrocks: When I wrote it I was sure it would be commented on, since I totally agree. Perhaps I should have commented on this.

It was like you're recommending at first, but since CsvLine requires that self.key_to_index in this object has been created already (circular dependency) I did it like this - rather than breaking up that dependency somehow. Decided to go with the hacky/non-readable solution in this case.

johandahlberg:
Ok. Great. I'm happy, so let me know if you'd like me to merge, or if you want to make any additional changes.

clarity_ext/domain/container.py
withrocks: Yes, so I believe these are not exactly the same use cases, since in your case you're yielding a generator in `outer_yield`, while in my case I'm looping over a generator and yielding its content. I think I've seen somewhere that your example should work as expected in Python 3.

However, I don't need to loop over the generator, as you point out. I'll need to do some fixes so I'll change this too right away.

withrocks:
I've pushed in two minor commits, parts of it has to do with your comments. Please merge it if there are no comments.

johandahlberg:
Looks good!

PR #56, opened 2016-11-03
Make pooled dilutions

GitEdvard:
Support driver file generation and updating udfs at pool creation steps. 

Implementation of function calculate_transfer_volumes is delegated out
to classes that is dedicated to the type of dilution:
* FixedVolumeCalc
* OneToOneConcentrationCalc
* PoolConcentrationCalc

The code in class PoolConcentrationCalc is new in this pull request. Here, source samples (representated by one end of a transfer
object) are first grouped so that there are a single destination pool within each group.

More information on how the calculation goes is written in the code comments.

### DIFF
--- a/clarity_ext/dilution_strategies.py
+++ b/clarity_ext/dilution_strategies.py
@@ -1,3 +1,25 @@
+# These values, residing in SingleTransfer objects, are calculated in
+# the these strategy classes (except FixedVolumeCalc)
+#
+# Variables to be used in driver file and udf update calculations:
+#   sample_volume,
+#   the volume in ul taken from the source sample into the destination
+#
+#   buffer_volume,
+#   the volume in ul taken from water/buffer/EB
+#
+# Variables to serve as basis of producing warning messages at validation.
+#   has_to_evaporate,
+#   Is true when the specified required concentration exceeds the source
+#   concentration
+#
+#   scaled_up,
+#   When a sample volume is less than the pipetting min volume of 2 ul,
+#   both sample volume and buffer volume will be scaled up, still rendering
+#   the right concentration but higher volume than desired.
+
+from itertools import groupby
+
 ROBOT_MIN_VOLUME = 2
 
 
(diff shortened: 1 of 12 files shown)


clarity_ext/dilution_strategies.py
```diff
@@ -0,0 +1,137 @@
+# These values, residing in SingleTransfer objects, are calculated in
```
withrocks: Looks good! I think dragging different strategies out of the service like this, following the single responsibility principle, and then injecting them into the service (as opposed to e.g. using class hierarchies) is definitely the way to go.

clarity_ext/dilution_strategies.py
```diff
@@ -0,0 +1,137 @@
...
+            try:
+                if scale_up_low_volumes is True and min_sample_volume < ROBOT_MIN_VOLUME:
+                    scale_factor = float(ROBOT_MIN_VOLUME/min_sample_volume)
+                    t.sample_volume *= scale_factor
+                    t.buffer_volume *= scale_factor
+                    t.scaled_up = True
+            except ZeroDivisionError:
+                # Zero sample volume indicates an error that should be caught
+                # later by validation
+                pass
```
withrocks: I'm always afraid of catch-all exception handlers. (I understand that in this case it will work, since we currently know how the class should be used - i.e. this call should be followed by validation. But if someone refactors this later, that might not be the case anymore.)

clarity_ext/domain/common.py
```diff
@@ -33,6 +33,9 @@ def _eq_rec(self, a, b, cache=[]):
                 continue
             elif any(b[key] is i for i in cache):
                 continue
+            elif a[key].__class__.__name__ == "MagicMock" and a[key].__class__.__name__ == "MagicMock":
```
withrocks: I definitely see the point with this as a short term solution to a problem, but I think this might give a false positive for other tests so I think we might want to refactor this later. I imagine other devs that haven't read this code writing tests that check for equality and don't know about this detail - they might get passed tests for domain objects that really are missing some crucial properties.

Without having dug deep, I would consider mocking the equality check instead. Then the responsibility for this resides in the tests only.

withrocks:
Merging. Added tickets per comments.

PR #59, opened 2016-11-08
Support generating files from templates

withrocks:
Add the TemplateExtension base class
Minor related fixes


johanherman:
Can merge this when the conflict is resolved.

clarity_ext/driverfile.py
```diff
@@ -63,9 +64,17 @@ def _save_file_locally(self, root):
...
+            content = self.specific_file_service.content()
+            if isinstance(content, basestring):
+                try:
+                    f.write(content)
+                except UnicodeEncodeError:
+                    f.write(content.encode("utf-8"))
+            else:
+                newline = self.extension.newline()
+                for line in content:
+                    f.write(line + newline)
```
withrocks: The method `content()` can return either a `basestring` (`str`/`unicode`) or a list of strings. If it's a list of string, it should use that newline attribute when writing to the file, othewise it should just write out the entire string.

withrocks:
Sounds good!

PR #60, opened 2016-11-08
Integrate with travis-ci

withrocks:
- Add .travis.yml
- Add a dummy .genologicsrc file. Even though it's not required by
  the unit tests, it's automatically loaded when the lib is imported.
  If the file doesn't exist, the application exits directly (rather
  than an error being thrown) so handling it would require monkey patching.
- Minor changes


johandahlberg:
Superb!

PR #65, opened 2016-12-06
Refactor ExtensionService.execute

withrocks:
- Move the logic to several individual methods
- Some cleanup

### DIFF
--- a/clarity_ext/cli.py
+++ b/clarity_ext/cli.py
@@ -7,6 +7,7 @@ from clarity_ext.extensions import ExtensionService
 from clarity_ext.tool.template_generator import TemplateNotFoundException, TemplateGenerator
 import os
 import yaml
+import time
 
 config = None
 logger = None
@@ -38,7 +39,6 @@ def validate(module):
     Validates the extension if there exists frozen data for it.
     Can use regex to match extensions.
     """
-    import time
     t1 = time.time()
     integration_svc = IntegrationTestService()
     validation_exceptions = integration_svc.validate(module, config)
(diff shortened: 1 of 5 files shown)


clarity_ext/cli.py
```diff
@@ -65,9 +65,31 @@ def extension(module, mode, args, cache):
...
+        if not config:
+            config = {
+                "test_root_path": "./clarity_ext_scripts/int_tests",
+                "frozen_root_path": "./clarity_ext_scripts/int_tests",
+                "exec_root_path": "."
+            }
+            logger.debug("Configuration not provided, using default: {}".format(config))
+
+        # Parse the run arguments list:
+        if args and isinstance(args, basestring):
```
johanherman: Couldn't you parse this more safely with e.g. `click`?

clarity_ext/extensions.py
```diff
@@ -73,164 +77,180 @@ def _run_path(self, args, module, mode, config):
...
+            use_cache = True
+        self._set_cache(use_cache)
+        instance = self._get_extension(module)(None)
+
+        self._prepare_runs(instance)  # Required to support certain tests
+        if not run_arguments_list:
+            run_arguments_list = self._gather_runs(module, True)
+
+        for run_arguments in run_arguments_list:
+            pid = run_arguments["pid"]
```
johanherman: Perhaps it just shows my ignorance for the arguments, but shouldn't you validate that these steps exist? If no `run_arguments_list` is passed into `run_test()` then `NoTestsFoundException` seem to be raised inside `_gather_runs()`, but if the user has supplied invalid `args` on the CLI level then I guess it would fail somewhere else, further down the rabbit hole (perhaps in the `ClaritySession` layer, but it is not immediately clear to me)?

withrocks: For some context: The arguments list actually just contains PIDs (step IDs). It's a legacy thing that it's called argument_list, I assume I had some extra parameters in there earlier. I didn't refactor it out now though to not do too much work in one PR.

So, when running tests, the PIDs are gathered from the extension being tested. If there isn't any test found, we raise an exception at this point. But generally, we do already have a PID (provided by the user through the command line, or by Clarity through their EPP configuration.) In that case, as you assumed, an exception is thrown when it tries to connect.

withrocks:
Please wait with merging this. I have other PRs that should go in at the same time, I had some tests breaking because the refactoring isn't complete, but I'd like to keep those changes in another PR to make it easier to code review.

withrocks:
Please wait with merging this, I'll add some stuff to the PR to make all tests work.

PR #67, opened 2016-12-13
Support freezing logs

withrocks:
- Log files should be frozen along with generated data
- Support test_mode exeuction of extensions. These ensure that
  generated data does remain constant between test runs.
- Support disabling commits while running extension integration tests
- Change the representation of a few domain objects, for log output
- Log UDF assignments so they can be compared between runs
- Add a handler for logging while testing. This handler only logs the
  log entries that were logged during execution of the extension
  and they land up in the run-test directory so they will eventually
  be frozen.
- Add a new execution method for extensions, test-fresh, which ignores
  the frozen cache
- Minor non-interface breaking related fixes

### DIFF
--- a/clarity_ext/cli.py
+++ b/clarity_ext/cli.py
@@ -8,9 +8,11 @@ from clarity_ext.tool.template_generator import TemplateNotFoundException, Templ
 import os
 import yaml
 import time
+from clarity_ext.extensions import ResultsDifferFromFrozenData
 
 config = None
-logger = None
+logger = logging.getLogger(__name__)
+log_level = None
 
 
 @click.group()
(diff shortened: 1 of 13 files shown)


clarity_ext/cli.py
```diff
@@ -82,20 +89,33 @@ def extension(module, mode, args, cache):
             key_values = (argument.split("=") for argument in separated)
             args = [{key: value for key, value in key_values}]
 
-        if mode == extension_svc.RUN_MODE_FREEZE:
+        validate_against_frozen = True # Indicates a run that should ignore the frozen directory
+        if mode == "test-fresh":
```
withrocks: Yes, no doubt. I'll fix it in the PR following this one, since it's a small fix.

PR #69, opened 2016-12-19
Lims 809 fix update fields

GitEdvard:
Update fields was previously not implemented (as I remember it)

Use two list variables holding transfers, replacing  the former self.transfers: 
self.split_row_transfers, any row containing more than 50 ul transfer volume is split up. This list is used when producing driver file. 
self.unsplit_transfers. This list is used when updating fields. 

Not yet validated:
* updating volume for samples that are scaled up (original pipetting volume was less than 2 ul)
* updating volume for samples that are evaporated.


withrocks:
Fixed the merge conflict and some tests that were failing after the merge and merged with a new pull request: https://github.com/Molmed/clarity-ext/pull/73

PR #70, opened 2016-12-19
Refactor handling of UDFs

withrocks:
- Remove tests involving the update matrix (test_integration_test_service.py)
  as the ability to not commit when testing means this is not required anymore.
- All UDFs are now accessed via `udf_<python name>` where that attribute is automatically
  available based on the udf name in Clarity. Ex.: If the artifact has the UDF "Current conc.",
  the object will have a magic attribute called udf_current_conc.
- Remove the static UDF_MAP
- Added some new domain objects, including the Process
- Developers do now fetch step UDFs via `self.context.current_step.udf_*`


clarity_ext/domain/udf.py
```diff
@@ -1,128 +1,215 @@
...
         """
         Maps a UDF name from Clarity to one that matches Python naming conventions
 
-        Example: 'Fragment Lower (bp)' => 'fragment_lower_bp'
+        The naming scheme may cause clashes, which would need to be resolved by the
+        caller (e.g. by ignoring UDFs that clash, or by throwing an exception)
+
+        Examples:
+          'Fragment Lower (bp)' => 'udf_fragment_lower_bp'
+          '% Total' => 'udf_total'
```
withrocks: Add a todo in jira on handling this? What about clashes against "Total". Would it make sense to map it to udf_percent_total?

clarity_ext/service/artifact_service.py
```diff
@@ -117,15 +117,28 @@ def all_shared_result_files(self):
 
     def update_artifacts(self, update_queue, whatif=False):
```
withrocks: Change whatif to commit.

PR #79, opened 2017-02-22
Fix an issue with newline when uploading

withrocks:
- Also fix issue with line endings when using templates.
  The templates should control the line endings.
- Minor refactoring


johandahlberg:
LGTM

PR #112, opened 2017-08-14
Fix warnings

withrocks:
- Warnings from transfers were not being logged
- Error/warning summary was not being shown

### DIFF
--- a/clarity_ext/extensions.py
+++ b/clarity_ext/extensions.py
@@ -272,10 +272,10 @@ class ExtensionService(object):
         context.cleanup()
         os.chdir(old_dir)
 
-        self.notify(instance.errors, instance.warnings,
-                    context.validation_service.error_count, context.validation_service.warning_count, context)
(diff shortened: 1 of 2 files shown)


johandahlberg:
I can't say I fully understand what this does - but from what I do understand it does look alright to me. However, if this is something that has regressed, perhaps it would be good to add a test for this to ensure it does not happen again? I'm not sure how difficult that would be however, so let me know what you think and then I can merge.

withrocks:
Excellent point. This area of the codebase still hasn't been cleaned up since in the prototype and I would definitely want tests on this level - particularly since this has regressed a few times. I've added a ticket on fixing this since I don't think I can do it right now (see LIMS-1361).

I've rebased with some minor changes. The extension should now show e.g.
```WARNING: clarity_ext_scripts.dilution.dna_dilution_start ran with 0 errors, 2 warnings```
when encountering errors, but:
```clarity_ext_scripts.dilution.dna_dilution_start ran successfully```
when there are no errors or warnings.

PR #118, opened 2017-09-27
1438 - Update pycharm template

withrocks:
The pycharm template now assumes that clarity-ext is contained within
the script project as a submodule, since that's how it's used here.


GitEdvard:
Ok, looks good

PR #128, opened 2017-11-07
Update the FileService for increased upload support

withrocks:
NOTE: This is a breaking change, in that it changes how we handle commits. Calls to commit are not required anymore. The extensions will automatically commit at the end. This has been the plan from the beginning.

Files can now be uploaded directly from the dev's workstation.

The files are not committed if the disable_commits flag is set

### DIFF
--- a/clarity_ext/extensions.py
+++ b/clarity_ext/extensions.py
@@ -260,8 +260,8 @@ class ExtensionService(object):
         elif issubclass(extension, GeneralExtension):
             try:
                 instance.execute()
-            except UsageError as e:
                 context.commit()
+            except UsageError as e:
                 # UsageErrors are deferred and handled by the notify method
                 # To support the case (legacy) if someone raises an error without adding it to the defer list,
                 # we add it to the instance too:
--- a/clarity_ext/extensions.py
+++ b/clarity_ext/extensions.py
@@ -131,7 +131,7 @@ class ExtensionService(object):
                 self._prepare_fresh_test(path)
 
             with utils.add_log_file_handler(os.path.join(path, "extensions.log"), False, ExtensionTestLogFilter()):
-                self._run(path, pid, module, artifacts_to_stdout, False, disable_context_commit=not commit, test_mode=True)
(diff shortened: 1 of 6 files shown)


clarity_ext/extensions.py
```diff
@@ -261,15 +261,14 @@ def _run(self, path, pid, module, artifacts_to_stdout, upload_files, disable_con
             try:
                 instance.execute()
             except UsageError as e:
+                context.commit()
```
PerLundmark: Should be in try clause?

GitEdvard:
I think it's a good idea to gather calls to the server under the commit command. When it comes to dilution, these parts were split up some distance from each other in code. 

I have some problem to understand your own comments though, "Calls to commit are not required anymore. The extensions will automatically commit at the end." ... The commit command is still there, isn't it?

clarity_ext/service/file_service.py
```diff
@@ -160,62 +171,61 @@ def upload_files(self, file_handle, files, stdout_max_lines=50):
...
+        self.logger.info("Queuing file '{}' for upload to the server".format(local_path))
+        file_name = os.path.basename(local_path)
+        if not file_name.startswith(artifact.id):
+            file_name = "{}_{}".format(artifact.id, file_name)
+        upload_path = os.path.join(self.upload_queue_path, file_name)
+        self.os_service.copy_file(local_path, upload_path)
+
+    def close_local_shared_files(self):
+        for f in self._local_shared_files:
+            f.close()
```
GitEdvard: Isn't it better to use context managers for this?

withrocks: Yes I think so too. But I believe (I wrote this a long time ago) that the idea here is that the caller should not have to know that things need to be closed, but that it would happen in the destructor of the fileservice object. But it would be cool to merge the ideas.

withrocks:
Right, I see your point about "not required". I meant that it's not required by the extension developers, as the framework now does it automatically in the end.

PR #140, opened 2017-11-27
Support pooling to 96 well plate

withrocks:
- Better support for seeing the original "virtual" transfer that should be executed.
- Includes related minor fixes (see individual commits)

### DIFF
--- a/clarity_ext/domain/container.py
+++ b/clarity_ext/domain/container.py
@@ -104,7 +104,7 @@ class Container(DomainObjectMixin):
     CONTAINER_TYPE_96_WELLS_PLATE = "96 well plate"
 
     def __init__(self, mapping=None, size=None, container_type=None,
-                 container_id=None, name=None, is_source=None, append_order=DOWN_FIRST):
+                 container_id=None, name=None, is_source=None, append_order=DOWN_FIRST, sort_number=0):
         """
         :param mapping: A dictionary-like object containing mapping from well
         position to content. It can be non-complete.
@@ -130,6 +130,8 @@ class Container(DomainObjectMixin):
         self.size = size
         self._append_iterator = None
         self.append_order = append_order
+        self.sort_number = sort_number
+        self.fixed_slot = None
 
     def append(self, artifact):
         """Adds this artifact to the next free position"""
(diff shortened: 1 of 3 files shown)


clarity_ext/service/dilution/service.py
```diff
@@ -431,6 +440,54 @@ def __repr__(self):
...
+
+        for transfer in transfers:
+            transfer.virtual_batch = self
+
+    def __getitem__(self, transfer):
+        """Returns the virtual transfer by the transfer"""
+        return self.virtual_transfers[transfer.target_location.artifact.id]
+
+    @staticmethod
+    def transfers_by_output(transfers):
```
GitEdvard: Maybe add a comment of what is the dict-key? It takes time to penetrate the return statement.

clarity_ext/service/dilution/service.py
```diff
@@ -431,6 +440,54 @@ def __repr__(self):
...
+        def group():
+            def group_key(transfer):
+                return transfer.target_location.artifact.id
+            s = sorted(transfers, key=group_key)
+            return {k: list(t) for k, t in groupby(s, key=group_key)}
+        return {k: VirtualTransfer(t) for k, t in group().items()}
+
+
+class VirtualTransfer(object):
```
GitEdvard: Maybe have a name that referers to pools?

GitEdvard:
Ok, ping me when you want to merge!

clarity_ext/domain/container.py
```diff
@@ -130,6 +130,8 @@ def __init__(self, mapping=None, size=None, container_type=None,
         self.size = size
         self._append_iterator = None
         self.append_order = append_order
+        self.sort_number = sort_number
```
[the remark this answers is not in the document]
withrocks: Yes, I think that's much better. I had though about sort_category or sort_weight but missed that sort_number will be easy to misunderstand.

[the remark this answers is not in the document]
withrocks: Went for sort_weight as I think that's very clearly indicating that it adds weight to the sort order.

clarity_ext/service/dilution/service.py
withrocks: :+1:

clarity_ext/service/dilution/service.py
withrocks: Yes, I think my tendency to try to be generic might be in the way here. I was though thinking that (further down the line) this would encapsulate the initial transfer that the RE does, even those that are not pools. I then go ahead and comment that it has to do with pools (only), which was a mistake. I hope I can clarify this in future comments.

PR #141, opened 2017-12-01
Support multiple files for local_shared_file

withrocks:


clarity_ext/service/file_service.py
```diff
@@ -79,12 +79,12 @@ def local_shared_file(self, file_name, mode='r', extension="", modify_attached=F
 
         # Ensure that the user is only sending in a "name" (alphanumerical or spaces)
         # File paths are not allowed
-        if not re.match(r"[\w ]+", file_name):
+        if not re.match(r"[\w ]+", file_handle):
```
PerLundmark: This would perhaps allow a match to the partial string ie match the text between for example the slashes in a path since it doesn't anchor the beginning and end of string?

withrocks: Good catch!

clarity_ext/service/file_service.py
```diff
@@ -147,15 +147,20 @@ def _queue(self, downloaded_path, artifact, file_prefix):
...
+        by_handle = [shared_file for shared_file in shared_files
+                     if shared_file.name == file_handle]
+
+        # Further filter
+        if file_name_contains is not None:
+            by_handle = [a for a in by_handle if file_name_contains in a.files[0].original_location]
+
+        if len(by_handle) != 1:
             files = ", ".join(map(lambda x: x.name, shared_files))
             raise SharedFileNotFound("Expected a shared file called '{}', got {}.\nFile: '{}'\nFiles: {}".format(
```
PerLundmark: Add file_name_contains to error output to make it easier to spot a problem?

PR #149, opened 2018-01-16
Support zipping files to a file handle

withrocks:
When uploading files through the file service, one can now
zip them together. In that case, only one file needs to be available
on the file handle. If zip_files is not specified, the previous logic
will work as expected, where each of the files is uploaded to a single
file handle (which must be configured specially on the step.)

Closes LIMS-1601

### DIFF
--- a/clarity_ext/extensions.py
+++ b/clarity_ext/extensions.py
@@ -578,7 +578,7 @@ class DriverFileExtension(GeneralExtension):
         return FileService.FILE_PREFIX_ARTIFACT_ID
 
 
-class SampleSheetExtension(DriverFileExtension):
+class SampleSheetExtension(GeneralExtension):
     """
     Provides helper methods for creating a CSV
     """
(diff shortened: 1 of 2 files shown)


GitEdvard:
I think it seem alright, but I think upload-files get a little bit bloated. The current function of upload-files is to associate files with artifacts, and it's not that obvious that this association will occur here. With more lines, it's harder to see that. 

Also, I think it may be a bit confusing when calling self._queue() for zip files, and self._upload-single for ordinary files. Is there a problem to call _upload single?

withrocks:
Thanks. I agree that it's getting bloated, moved the zip stuff into a separate method and added some comments.

Using `_upload_single` would require refactoring of that method, because it expects a string, `content`, that's written to disk and then queued, so that method is really just a wrapper around queuing when you've got a string rather than a file. This feature is kind of a quick fix so I didn't want to refactor too much, but I think we should definitely clear this up later!

PR #151, opened 2018-02-19
Skip updating the step log if any sample is under review

withrocks:
Otherwise the script will fail in that case

### DIFF
--- a/clarity_ext/service/file_service.py
+++ b/clarity_ext/service/file_service.py
@@ -265,7 +265,13 @@ class FileService:
                                              if shared_file.id == artifact_id])
                     local_file = os.path.join(self.upload_queue_path, artifact_id, file_name)
                     self.logger.info("Uploading file {}".format(local_file))
-                    self.session.api.upload_new_file(artifact.api_resource, local_file)
+                    try:
+                        self.session.api.upload_new_file(artifact.api_resource, local_file)
+                    except requests.HTTPError as e:
+                        if "UNDER_REVIEW" in str(e):
+                            self.logger.error("Not able to upload step log as some of the samples are in review")
+                        else:
+                            raise e
 
     def _split_file_name(self, name):
         m = re.match(self.SERVER_FILE_NAME_PATTERN, name)


GitEdvard:
If you have tested this and it's OK, than I think it's OK. But..., I can't really understand why the step log would hinder the lims to terminate the step. When looking at the code, it could be any file in the step that causes problem, and it feels more probable that it would be a file associated with one of those samples. 

Another thing came into my mind here. Is the files that are saved under upload_quue and downloaded ever deleted? In file_service, they are just unlinked in lims-server, the local files are not removed.

clarity_ext/service/file_service.py
```diff
@@ -265,7 +265,13 @@ def commit(self, disable_commits):
                                              if shared_file.id == artifact_id])
                     local_file = os.path.join(self.upload_queue_path, artifact_id, file_name)
                     self.logger.info("Uploading file {}".format(local_file))
-                    self.session.api.upload_new_file(artifact.api_resource, local_file)
+                    try:
+                        self.session.api.upload_new_file(artifact.api_resource, local_file)
+                    except requests.HTTPError as e:
+                        if "UNDER_REVIEW" in str(e):
+                            self.logger.error("Not able to upload step log as some of the samples are in review")
```
GitEdvard: I'm little confused here. If it's step log that is not working, then message to logger will not be shown to user (ok, from API though...). 

If it by any chance it's not the step log that causes problem, then this message might or might not be uploaded to server, depending on when step log is uploaded and when problem-file is uploaded.

GitEdvard:
Ok, I understand this is a little bit urgent to get in. Just tell me to merge, despite the large amount of comments. ...

withrocks:
Thanks for the comments. I'm certain that this is the problem here even if it looks a bit weird. The thing is that the step log is connected to the samples, it's considered a shared result file for all of them, and Clarity has the feature (which makes sense I guess) that nothing should be changed that relates to a sample under review.

The error log I'm writing to will show up in our logs, i.e. those that are available on the server, and they will generally never fail. They will turn up for us in the future in ELK but can currently at least be read on the server.

Regarding the question about if files are deleted, they are, I believe, not. Just unlinked as you mentioned.

PR #156, opened 2018-04-12
Add method to check file extension

GitEdvard:
Purpose:
User may easily by mistake upload a .xlsx file when either .csv or xml is required. Show a clear error message for this. 

In context.local_shared_file(), if the searched file is csv or xml,
check that the found artifact has a file with the right extension, i.e.
.csv or .xml. If not, cast exception with proper message.

### DIFF
--- a/clarity_ext/service/file_service.py
+++ b/clarity_ext/service/file_service.py
@@ -266,6 +266,22 @@ class LocalSharedFileProvider:
         return self._local_shared_file(artifact, filename, mode=mode, extension=extension,
                                        modify_attached=modify_attached)
 
+    def check_file_extension(self, clarity_file_handle, required_extension=None, filename_contains=None):
+        """
+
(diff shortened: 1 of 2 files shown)


clarity_ext/service/file_service.py
```diff
@@ -266,6 +266,15 @@ def search_or_create(self, file_handle, mode='ab', extension="",
         return self._local_shared_file(artifact, filename, mode=mode, extension=extension,
                                        modify_attached=modify_attached)
 
+    def check_file_extension(self, file_handle, required_extension=None, filename_contains=None):
+        artifact = self._artifact_by_name(file_handle, filename=filename_contains)
+        file_name = os.path.basename(artifact.file_name)
+        _, actual_extension = os.path.splitext(file_name)
+        if not required_extension in file_name:
```
PerLundmark: Wouldn't it be better to check the extension, to avoid false positives? Since you split it out anyway?

PR #164, opened 2019-03-27
 1893 dilute to tubes

GitEdvard:
Prepare clarity-ext to be able for dilution to tubes.

Implementation:
Add class TubeRackPositioner. Temporarily create  tube racks and place artifacts in it, in the order of the  transfers list. Therefore, sort transfers prior to placing tubes to tube racks. Also, update container with type TUBE. 

Refactorings:
method container_sort_key is moved from clarity-snpseq to here.

### DIFF
--- a/clarity_ext/service/dilution/service.py
+++ b/clarity_ext/service/dilution/service.py
@@ -1,7 +1,9 @@
 import abc
 import copy
 import logging
+import re
 from itertools import groupby
+from itertools import chain
 from itertools import izip_longest
 import collections
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


clarity_ext/service/dilution/service.py
```diff
@@ -532,6 +549,90 @@ def output_position_sort_key(transfer):
...
+
+        self.tube_racks[-1].set_well(well_pos=self.current_well_pos, artifact=well.artifact)
+
+    @property
+    def last_populated_well(self):
+        return self.tube_racks[-1].wells[self.current_well_pos]
+
+    def _convert_to_coordinates(self, position_index):
+        rowind = position_index % self.size.height
+        colind = position_index // self.size.height
```
withrocks: Should it be `self.size.height` in both cases?

clarity_ext/service/dilution/service.py
```diff
@@ -532,6 +549,90 @@ def output_position_sort_key(transfer):
...
+        return rowind, colind
+
+    def _convert_to_position(self, row_ind, col_ind):
```
withrocks: I wonder if this function should return a string or `ContainerPosition`. It's not used at the moment, so difficult to be sure.

clarity_ext/service/dilution/service.py
```diff
@@ -632,6 +733,14 @@ def source_container_name(transfer_location):
     def target_container_name(transfer_location):
         return "END{}".format(transfer_location.container_pos)
 
+    @abc.abstractmethod
```
withrocks: If you would introduce a second variable in the settings (see the other PR) I think you could skip this abstract method.

clarity_ext/service/dilution/service.py
withrocks: Good to know :)

withrocks:
Yeah, I agree that it's gotten long. I was thinking that one could leave it out of that and just use different settings like in this: 

https://gitlab.snpseq.medsci.uu.se/lims/clarity-snpseq/commit/dc1cfa6a2b9835ecffafd0a612305e73c0d6a4e0#db76b23841262b4ce98b82c68e2f81ae0b772fcb_292_292

But that has it's downsides also. I guess that adding it to the session is a better solution.

PR #165, opened 2019-04-11
Ext 1892 loop bug

GitEdvard:
Fix bug with symptoms:
User make a dilution with a lot of samples, that requires more than one intermediate plate.

Implementation:
Update method get_temporary_container(), so that it provides a new temp plate for each target plate containing looped samples. Previously there was only one temp plate. 

Refactorings:
DilutionSession.get_temporary_container() is moved to separate class (maintaining a state). get_temporarily_container() also have two new input arguments, start_id and container_template. Container template is used when diluting to tube racks, in which case the intermediate plate should still be a 96 well plate (not dimensions of the tube rack).

### DIFF
--- a/clarity_ext/service/dilution/service.py
+++ b/clarity_ext/service/dilution/service.py
@@ -4,7 +4,6 @@ import logging
 import re
 from itertools import groupby
 from itertools import chain
-from itertools import izip_longest
 import collections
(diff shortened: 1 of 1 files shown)


withrocks:
Looks good to me!

PR #166, opened 2019-04-15
1894 pooling to tubes

GitEdvard:
Purpose:
Adapt TubeRackPositioner for pools, so that each pool retreive exactly one position in tube rack. 

Implementation:
TubeRackPositioner, which give output artifacts a position in a tube rack, get one call for each transfer. In case of pooling, there are multiple calls for the same pool. Prevent that the same pool are placed at multiple places in tube rack, by searching for artifacts that are previously added.

Add output_position_pre_batching to SortStrategy. In DilutionSession, sort transfers if a sort strategy is set, before they are treated by handlers. 

Refactorings:
* Change variable name, SharedRobotSettings.max_pipette_vol_for_row_split --> max_destination_volume.
* Change variable name, RobotSettings.sort_strategy --> robotfile_sort_strategy

### DIFF
--- a/clarity_ext/service/dilution/service.py
+++ b/clarity_ext/service/dilution/service.py
@@ -213,6 +213,7 @@ class DilutionSession(object):
         original_containers = set()
         original_containers.update([pair.input_artifact.container for pair in pairs])
         original_containers.update([pair.output_artifact.container for pair in pairs])
+
         for original_container in original_containers:
             containers[original_container.id] = copy.copy(original_container)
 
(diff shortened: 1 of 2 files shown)


clarity_ext/service/dilution/service.py
```diff
@@ -122,13 +122,19 @@ def evaluate_transfer_route(self, transfer, transfer_handlers):
...
-            self.max_pipette_vol_for_row_split = robotsettings.max_pipette_vol_for_row_split_plate
+            self.max_destination_volume = robotsettings.max_destination_volume_plate
 
     def _get_destination_container_type(self, transfers):
         types = list(set([t.target_location.artifact.container.container_type for t in transfers]))
-        return types[0]
+        return utils.single(types)
+
+    def _sorted_transfers(self, original_transfers):
+        sort_strategy = self.dilution_settings.tube_placement_sort_strategy
```
withrocks: I wonder if `tube_placement_sort_strategy` shouldn't have a more generic name (perhaps just sort_strategy) since it's being addressed in code that will work for tubes as well as plates.

GitEdvard: I wanted to distinguish the sort strategy for placing tubes from the sort strategy for sorting the robotfile, that is used later in the dilution code. (robotfile sorting could not be used here because all handlers must be exectued first). I wonder if it could be okay to keep the tube-placement-sort-strategy? When configuring the dilution-settings, I think its fairly explicit what it's about.

withrocks: Ahh I see!

PR #167, opened 2019-04-23
1540, Add reagent type to the domain model

GitEdvard:
Purpose:
Add reagent-type to the domain model, so that the attributes are written out explicitly. Also, initialize files for Errors and Warnings in the context class. 

Implementation:
reagent-type needs a repository, that is added.


withrocks:
Looks good to me!

PR #168, opened 2019-04-26
Ext 1925 pool with input tubes

GitEdvard:
Purpose:
1) Add tuberack as a new container type. This is used in clarity-snpseq to check if tuberack should be shown in metadatafile.
2) Configurable name prefix for tuberacks, either destination tuberack, or source tuberack.


withrocks:
Looks good!

PR #171, opened 2019-06-14
1959 convert index pos for hamilton

GitEdvard:
Purpose:
Move some code from clarity-snpseq-scripts (index file generation) to the framework. This code was previously only used in one script, but is now shared by two scripts.


withrocks:
Looks good to me!

PR #173, opened 2019-07-03
Ext 1970 remove duplication

GitEdvard:
Purpose:
Create a slim test framework for parsing scripts.

Implementation:
* Instantiate ExtensionContext without connecting to network, only using fake entities
* Configure ExtensionContext with a number of fake artifact pairs (input & output)
* Configure each fake artifact pair with a number of udfs required by the script

### DIFF
--- a/clarity_ext/repository/step_repository.py
+++ b/clarity_ext/repository/step_repository.py
@@ -100,16 +100,6 @@ class StepRepository(object):
             raise NotImplementedError(
                 "Generation type {} is not implemented".format(output_gen_type))
 
-        if isinstance(output, ResultFile):
-            output.output_type = Artifact.OUTPUT_TYPE_RESULT_FILE
-        elif isinstance(output, Analyte):
-            output.output_type = Artifact.OUTPUT_TYPE_ANALYTE
-        elif isinstance(output, SharedResultFile):
-            output.output_type = Artifact.OUTPUT_TYPE_SHARED_RESULT_FILE
-        else:
-            raise NotImplementedError(
-                "Output type {} is not implemented".format(output.__class__.__name__))
-
         # Add a reference to the other object for convenience:
         input.output = output
         output.input = input
--- a/clarity_ext/utility/testing_parse_scripts/builders.py
+++ b/clarity_ext/utility/testing_parse_scripts/builders.py
@@ -1,0 +1,122 @@
+from mock import MagicMock
+from StringIO import StringIO
+from io import BytesIO
+from clarity_ext.service.file_service import Csv
+from clarity_ext.service.file_service import FileService
+from clarity_ext.service.process_service import ProcessService
+from clarity_ext.context import ExtensionContext
+from clarity_ext.service.artifact_service import ArtifactService
+from clarity_ext.domain.process import Process
+from clarity_ext.domain.user import User
+from clarity_ext.domain.udf import UdfMapping
+from clarity_ext.utility.testing_parse_scripts.fake_artifact_factory import FakeArtifactFactory
+
+
+class ContextBuilder:
+    """
+    Add entities to repositories held by context
+    E.g. shared files, artifacts, analytes.
+    """
+    def __init__(self):
+        self.step_repo = FakeStepRepo()
+        self.logger = FakeLogger()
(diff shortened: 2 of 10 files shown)


clarity_ext/repository/step_repository.py
```diff
@@ -100,16 +100,6 @@ def _wrap_input_output(self, input_info, output_info, container_repo, process_ty
...
                 "Generation type {} is not implemented".format(output_gen_type))
 
-        if isinstance(output, ResultFile):
-            output.output_type = Artifact.OUTPUT_TYPE_RESULT_FILE
-        elif isinstance(output, Analyte):
-            output.output_type = Artifact.OUTPUT_TYPE_ANALYTE
-        elif isinstance(output, SharedResultFile):
-            output.output_type = Artifact.OUTPUT_TYPE_SHARED_RESULT_FILE
-        else:
-            raise NotImplementedError(
```
johandahlberg: To make sure that this retains backwards compatibility, perhaps the `NotImplementedError` should be thrown by the base class that the other classes implement?

clarity_ext/utility/testing_parse_scripts/builders.py
```diff
@@ -0,0 +1,122 @@
...
+    def with_analyte_pair(self, input, output):
+        self.step_repo.add_analyte_pair(input, output)
+
+    def with_mocked_local_shared_file(self, filename, contents):
+        monkey = LocalSharedFilePatcher()
+        monkey.cache[filename] = contents
+        self.context.local_shared_file = monkey.local_shared_file
+
+    def _create(self):
+        session = MagicMock()
```
johandahlberg: When creating the mock objects I think it might be worth using `create_autospec` (https://docs.python.org/3/library/unittest.mock.html#unittest.mock.create_autospec), which will decrease the likelyhood that the tests fail for the wrong reason. I.e. when just using a `MagicMock`, every method called will return an instance of `MagicMock`, even if that method is not defined on the object it tries to mock. If autospecing is used, it will thrown an exception if a method that does not exist on the specified object is called.

clarity_ext/utility/testing_parse_scripts/builders.py
```diff
@@ -0,0 +1,122 @@
...
+        file_repository = MagicMock()
+        artifact_service = ArtifactService(self.step_repo)
+        os_service = MagicMock()
+        current_user = None
+        file_service = FileService(
```
johandahlberg: There seems to be a mix of Mocks and actual implementations here, e.g. `FileService` and `ProcessService` use the actual implementations. Is there any particular reason for this?

clarity_ext/utility/testing_parse_scripts/builders.py
```diff
@@ -0,0 +1,122 @@
...
+        self.pair = None
+
+    def create(self):
+        pair = self.artifact_repo.create_pair(
+            pos_from=None, pos_to=None, source_id=None, target_id=self.target_id,
+            target_type=self.target_type)
+        pair.output_artifact.udf_map = UdfMapping(self.output_udf_dict)
+        self.pair = pair
+
+    def with_target_id(self, target_id):
```
johandahlberg: I not sure what convention has been using in clarity-ext, but normally I'd expect these types of methods to be called `set_*`. I don't object to this, I just wanted to check that there is a reasoning behind it.

GitEdvard:
Thanks for the comments. I've fixed most of them, and also removed the extra __init__ class.

clarity_ext/utility/testing_parse_scripts/builders.py
johandahlberg: :+1:

clarity_ext/utility/testing_parse_scripts/builders.py
johandahlberg: Cool. I think that makes sense, and avoids the problem I was trying to adress. :+1:

clarity_ext/utility/testing_parse_scripts/builders.py
johandahlberg: That makes sense!

PR #179, opened 2020-02-17
lims-0, Fix target volume at dilution

GitEdvard:
Purpose:

Fix bug that occurs at scale up in dilutions. When scale up occurs, the new sample volume is still registered as the user value of target volume, instead of the scaled up volume. 

Implementation:
Target volume (the volume for the sample leaving the dilution step)  is now based on the sum of pipetting volumes, instead of user value for target volume. The user value of target volume may be
scale up.

### DIFF
--- a/clarity_ext/service/dilution/service.py
+++ b/clarity_ext/service/dilution/service.py
@@ -422,8 +422,9 @@ class SingleTransfer(object):
 
     @property
     def update_info(self):
+        end_volume = self.pipette_sample_volume + self.pipette_buffer_volume
         return UpdateInfo(target_conc=self.target_conc if self.should_update_target_conc else None,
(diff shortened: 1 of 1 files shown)


clarity_ext/service/dilution/service.py
```diff
@@ -422,8 +422,9 @@ def updated_source_vol(self):
 
     @property
     def update_info(self):
+        end_volume = self.pipette_sample_volume + self.pipette_buffer_volume
```
matrulda: This looks like it's the same as the `pipette_total_volume` property, can you use that instead?

GitEdvard:
I reverted this commit, and instead fixed the problem in clarity-snpseq. See comments there.

PR #184, opened 2020-03-06
Add property on SingleTransfer

GitEdvard:


clarity_ext/service/dilution/service.py
```diff
@@ -410,6 +410,10 @@ def is_last_transfer(self):
         """If this transfer represent the last action against the target"""
         return self.batch == 'default' or self.batch == 'evaporate2'
 
+    @property
```
withrocks: I'm merging this since it's fixing a bug. But we might want to keep business logic out of the data contract and have the handlers handle that, right? The reason I think this is that our cases get so complex that this might be true in some scenarios and not in others, so running a particular series of handlers might lead this to be true, while running another series of handlers makes it false.

PR #197, opened 2020-05-05
Add utility method to list filenames

GitEdvard:
Purpose:
Utility method to list file names within a file handle.

### DIFF
--- a/clarity_ext/service/file_service.py
+++ b/clarity_ext/service/file_service.py
@@ -256,6 +256,9 @@ class FileService:
                 raise NotImplementedError("Type not supported")
         return full_path
 
+    def list_filenames(self, file_handle):
+        return self.local_shared_file_provider.list_filenames(file_handle)
+
 
 class LocalSharedFileProvider:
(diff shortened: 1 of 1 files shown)


clarity_ext/service/file_service.py
```diff
@@ -267,6 +270,14 @@ def __init__(self, file_service, file_repo, artifact_service, downloaded_path, o
         self.should_cache = should_cache
         self.logger = logger
 
+    def list_filenames(self, file_handle):
+        shared_files = self.artifact_service.shared_files()
+        by_handle = [
+            shared_file for shared_file in shared_files
+            if shared_file.name == file_handle
+        ]
+        return [a.files[0].original_location for a in by_handle]
```
withrocks, line 279: Do you know if `original_location` returns the path and the file name? I don't remember this one (the name indicates only a path to me.)

GitEdvard: I checked now (in integration tests). It seems to be only the file name.

withrocks: Ok!

PR #199, opened 2020-06-02
Changes made during the covid project

withrocks:
This PR includes all the changes that were made in the Covid project. These should be safe for SNP&SEQ, but I recommend that we do a full validation of our workflows when we start using this in that project.

### DIFF
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from clarity_ext.service.dilution.service import DilutionService
 from clarity_ext import UnitConversion
 from clarity_ext.repository import ClarityRepository, FileRepository
@@ -63,6 +64,7 @@ class ExtensionContext(object):
         self.disable_commits = disable_commits
         self._calls_to_commit = 0
         self.validation_results = list()
+        self.start = datetime.now()
 
     @staticmethod
     def create(step_id, test_mode=False, uploaded_to_stdout=False, disable_commits=False):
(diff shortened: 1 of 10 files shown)


GitEdvard:
There is a covid branch in genologics repo as well. Shouldn't we merge that one before this one?

clarity_ext/domain/process.py
```diff
@@ -4,10 +4,11 @@
...
     """Represents a Process (step)"""
 
-    def __init__(self, api_resource, process_id, technician, udf_map, ui_link):
+    def __init__(self, api_resource, process_id, technician, udf_map, ui_link, instrument):
         super(Process, self).__init__(api_resource, process_id, udf_map)
         self.technician = technician
         self.ui_link = ui_link
+        self.instrument = instrument
 
     @staticmethod
```
GitEdvard, line 13: Currently, we don't use instrument on step level at snpseq. Should we include this anyway?

clarity_ext/context.py
```diff
@@ -263,12 +263,13 @@ def update(self, obj):
 
     def commit(self):
         """Commits all objects that have been added via the update method, using batch processing if possible"""
-        self._calls_to_commit += 1
-        if self._calls_to_commit > 1:
-            self.logger.warning("Commit called more than once. It's not necessary to call commit explicitly anymore.")
```
GitEdvard, line 259: This warning is perhaps still relevant in snpseq?

withrocks: It might be, but it has no effect to remove it either.

clarity_ext/domain/process.py
withrocks: I think it makes sense since this is available on this level in Clarity. It doesn't mean we need to use it there.

PR #202, opened 2020-10-08
Lazy load samples on an artifact

withrocks:
Don't load all samples when creating the domain object version of an
Artifact. This was the reason for extremely long running scripts
including pools, as we are not batch fetching the samples.

A further improvement would batch fetch also (see the samples property
on the Aliquot).

### DIFF
--- a/clarity_ext/domain/aliquot.py
+++ b/clarity_ext/domain/aliquot.py
@@ -10,10 +10,19 @@ class Aliquot(Artifact):
     of the original for example. Or, in the case of ResultFile, only a measurement of the original.
     """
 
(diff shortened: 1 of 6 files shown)


matrulda (approved):
LGTM!

matrulda:
@withrocks Can't merge since travis hasn't run yet. I can't really see what is happening there, can you?

withrocks:
@matrulda Right! It seems like it's in a limbo. I'll take a look.

withrocks:
I ran the test manually and it's passing. I'll force merge this.

PR #204, opened 2020-10-15
Create samples in a batch

withrocks:


clarity_ext/service/clarity_service.py
```diff
@@ -102,39 +102,44 @@ def get_project_by_name(self, project_name):
...
                     self.session.api,
-                    container=container_res,
+                    container_res,
                     position=repr(well.position),
                     name=sample.name,
                     project=sample.project.api_resource,
                     udfs=sample.udf_map.to_dict())
+            instances.append(instance)
+
+        instances = entities.Sample.batch_create(self.session.api, instances)
```
GitEdvard: I'm not fond of having variable "instances" change its meaning here. I would prefer to have two variables with something like in_memory_instances, and created_instances.

withrocks: Sounds good!

GitEdvard (approved):

PR #205, opened 2020-10-15
A couple of bug fixes

withrocks:
* We need a setter for samples in the Aliquot class as they are being
explicitly set in some tests
* Instrument should default to None so we don't need to change the
instantiation all over the place


GitEdvard (approved):
Looks good!

PR #207, opened 2021-01-28
1973 write script name to log

kleppari:
-  Add script name to log at start and end when exec scripts
-  Add commit feature to template

### DIFF
--- a/clarity_ext/extensions.py
+++ b/clarity_ext/extensions.py
@@ -260,7 +260,6 @@ class ExtensionService(object):
         instance = extension(context)
         context.logger.log("Start running script '{}'".format(module))
         # We add the config in this way (not via __init__) as a quick fix.
-        # Test
         instance.config = config
         try:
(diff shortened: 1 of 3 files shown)


clarity_ext/extensions.py
```diff
@@ -258,7 +258,9 @@ def _run(self, config, path, pid, module, artifacts_to_stdout, disable_context_c
                                           disable_commits=disable_context_commit,
                                           uploaded_to_stdout=artifacts_to_stdout)
         instance = extension(context)
+        context.logger.log("Start running script '{}'".format(module))
         # We add the config in this way (not via __init__) as a quick fix.
+        # Test
```
GitEdvard: This line looks misplaced.

GitEdvard (approved):

clarity_ext/extensions.py
kleppari: Will remove this

PR #208, opened 2021-03-16
Dev 1001, set qc flags from clarity-ext instead of lab logic

GitEdvard:
Purpose:
Prepare to set qc flags on artifacts from clarity-ext scripts instead of from lab logic.

Details:
qc-flag happen to be included in the genologics layer, so in clarity-ext the qc-flag property is just fetched from genologics layer. Since the qc-flag is an attribute directly on an artifact (rather than an udf), the qc-flag has to be addressed directly by name in domain classes when fetching/updating.


withrocks (approved):

withrocks:
Looks great!

PR #209, opened 2021-03-26
Develop 1171, fix qc flag bug

GitEdvard:
This PR can be merged when
* Commit f64e022 is re-introduced (because it has been reverted as now)
* Clarity-ext has been converted to python 3
* it is rebased according to python 3 conversion.

Purpose:
Fetch all artifacts without the state argument in the uri, so that artifacts are fetched as presented at UI. This will prevent some bugs and hack code.

As a precaution, also fetch artifacts in the previous way, including the state argument, then make a comparison of stateful and stateless xml represenation of the artifacts. In this way, each fetched artifact is validated. This validation is thought to exist for a limited period of time, because there is a performance loss in fetching the double amount of data for each step. 

Any such validation error are to be pushed into the step log and to the underlying system log. 

More information may be found at https://snpseq.atlassian.net/browse/DEVELOP-1177

Details:
A validation error from xml discordance can't be pushed right away, because when it happens the step log is not yet instantiated.

I made some refactoring in how the commit works after an error occurred.
1. Ensure only step log, errors and warnings are committed after an error occurred.
2. Any type of unexpected error is pushed to the step log and the underlying system log
Up to now, files may have been committed even after an error occurred, and in some cases also analytes. This is of course not good, although (as I understand it) the lab has the procedure to completely start a step from scratch after an error.

### DIFF
--- a/clarity_ext/extensions.py
+++ b/clarity_ext/extensions.py
@@ -16,6 +16,8 @@ from clarity_ext.service import ArtifactService, FileService
 from clarity_ext.utility.integration_test_service import IntegrationTest
 from clarity_ext.service.dilution.index_generation import ConfigValidator
 from clarity_ext.service.dilution.index_generation import ConfigParser
+from clarity_ext.domain.validation import ValidationException
+from clarity_ext.domain.validation import ValidationType
 from jinja2 import Template
 import time
 import random
@@ -258,10 +260,11 @@ class ExtensionService(object):
                                           disable_commits=disable_context_commit,
                                           uploaded_to_stdout=artifacts_to_stdout)
         instance = extension(context)
-        context.logger.log("Start running script '{}'".format(module))
-        # We add the config in this way (not via __init__) as a quick fix.
-        instance.config = config
         try:
+            context.logger.log("Start running script '{}'".format(module))
+            context.validate_xml_concordance()
+            # We add the config in this way (not via __init__) as a quick fix.
+            instance.config = config
             if issubclass(extension, DriverFileExtension):
                 context.file_service.upload(instance.shared_file(), instance.filename(), instance.to_string(),
                                             instance.file_prefix())
@@ -272,19 +275,28 @@ class ExtensionService(object):
             context.logger.log("Finished running script '{}'".format(module))
             context.commit()
         except UsageError as e:
-            # Commit in order to upload step log
-            # Ordinary files and objects shouldn't have been moved to upload queue,
-            # since the scipt have been interrupted by the usage error.
-            context.commit()
-            # UsageErrors are deferred and handled by the notify method
-            # To support the case (legacy) if someone raises an error without adding it to the defer list,
-            # we add it to the instance too:
-            if len(instance.errors) == 0:
-                instance.errors[e] = list()
+            self._commit_step_log(context, instance, e)
+        except Exception as e:
+            # All other unexpected exceptions should also be included in the step log
+            # as well as the underlying server log
+            error = ValidationException(str(e), ValidationType.ERROR)
+            context.validation_service.handle_single_validation(error)
+            self._commit_step_log(context, instance, e)
+
         os.chdir(old_dir)
 
         self.notify(instance.errors, instance.warnings, context.validation_service.error_count,
                     context.validation_service.warning_count, context, module)
+
+    def _commit_step_log(self, context, instance, error):
+        # Commit in order to upload step log
+        # Ordinary files and objects are not updated because the update queue are cleared
+        context.commit_step_log_only()
+        # UsageErrors are deferred and handled by the notify method
+        # To support the case (legacy) if someone raises an error without adding it to the defer list,
+        # we add it to the instance too:
+        if len(instance.errors) == 0:
+            instance.errors[error] = list()
 
     def notify(self, user_errors, user_warnings, other_errors_count, other_warnings_count, context, module):
         """Notifies the user of errors and warnings during execution. user_errors and user_warnings
--- a/clarity_ext/extensions.py
+++ b/clarity_ext/extensions.py
@@ -275,23 +275,22 @@ class ExtensionService(object):
             context.logger.log("Finished running script '{}'".format(module))
             context.commit()
         except UsageError as e:
-            self._commit_step_log(context, instance, e)
+            context.commit_step_log_only()
+            self._ensure_error_on_instance(instance, e)
         except Exception as e:
             # All other unexpected exceptions should also be included in the step log
             # as well as the underlying server log
             error = ValidationException(str(e), ValidationType.ERROR)
             context.validation_service.handle_single_validation(error)
-            self._commit_step_log(context, instance, e)
+            context.commit_step_log_only()
+            self._ensure_error_on_instance(instance, e)
 
         os.chdir(old_dir)
 
         self.notify(instance.errors, instance.warnings, context.validation_service.error_count,
                     context.validation_service.warning_count, context, module)
 
-    def _commit_step_log(self, context, instance, error):
-        # Commit in order to upload step log
-        # Ordinary files and objects are not updated because the update queue are cleared
-        context.commit_step_log_only()
+    def _ensure_error_on_instance(self, instance, error):
         # UsageErrors are deferred and handled by the notify method
         # To support the case (legacy) if someone raises an error without adding it to the defer list,
         # we add it to the instance too:
--- a/clarity_ext/repository/step_repository.py
+++ b/clarity_ext/repository/step_repository.py
@@ -1,6 +1,5 @@
+from collections import namedtuple
 from clarity_ext.domain.artifact import Artifact
-from clarity_ext.domain.analyte import Analyte
-from clarity_ext.domain.result_file import ResultFile
 from clarity_ext.domain.shared_result_file import SharedResultFile
 from clarity_ext.repository.container_repository import ContainerRepository
 from clarity_ext.domain.user import User
@@ -44,16 +43,7 @@ class StepRepository(object):
         instead.
         """
         input_output_maps = self.session.current_step.api_resource.input_output_maps
-        artifact_keys = set()
-        for input, output in input_output_maps:
-            artifact_keys.add(input["uri"])
-            artifact_keys.add(output["uri"])
-
-        artifacts = self.session.api.get_batch(artifact_keys)
-        artifacts_by_uri = {artifact.uri: artifact for artifact in artifacts}
-        for input, output in input_output_maps:
-            input['uri'] = artifacts_by_uri[input['uri'].uri]
-            output['uri'] = artifacts_by_uri[output['uri'].uri]
+        wrappable_pairs = WrappablePairs(self.session, input_output_maps, WrappablePairs.MODE_STATELESS)
 
         # Artifacts do not contain UDFs that have not been given a value. Since the domain objects returned
         # must know all UDFs available, we fetch them here:
@@ -67,9 +57,14 @@ class StepRepository(object):
         # that we create only one artifact domain object in this case:
         outputs_by_id = dict()
         container_repo = ContainerRepository()
-        for input_res, output_res in input_output_maps:
+        for genologics_input, genologics_output, output_generation_type in wrappable_pairs:
             input, output = self._wrap_input_output(
-                input_res, output_res, container_repo, process_type)
+                genologics_input,
+                genologics_output,
+                container_repo,
+                process_type,
+                output_generation_type
+            )
 
             if output.id in outputs_by_id:
                 output = outputs_by_id[output.id]
@@ -77,16 +72,14 @@ class StepRepository(object):
             outputs_by_id[output.id] = output
         return ret
 
-    def _wrap_input_output(self, input_info, output_info, container_repo, process_type):
+    def _wrap_input_output(self, input_resource, output_resource, container_repo,
+                           process_type, output_generation_type):
 
         # Create a map of all containers, so we can fill in it while building
         # domain objects.
 
         # Create a fresh container repository. Then we know that only one container
         # will be created for each object in a call to this method
-        input_resource = input_info["uri"]
-        output_resource = output_info["uri"]
-        output_gen_type = output_info["output-generation-type"]
         input = self._wrap_artifact(
             input_resource,
             container_repo,
@@ -97,17 +90,17 @@ class StepRepository(object):
         output = self._wrap_artifact(
             output_resource,
             container_repo,
-            gen_type=output_gen_type,
+            gen_type=output_generation_type,
             is_input=False,
             process_type=process_type)
 
-        if output_gen_type == "PerInput":
+        if output_generation_type == "PerInput":
             output.generation_type = Artifact.PER_INPUT
-        elif output_gen_type == "PerAllInputs":
+        elif output_generation_type == "PerAllInputs":
             output.generation_type = Artifact.PER_ALL_INPUTS
         else:
             raise NotImplementedError(
-                "Generation type {} is not implemented".format(output_gen_type))
+                "Generation type {} is not implemented".format(output_generation_type))
 
         # Add a reference to the other object for convenience:
         # TODO: There are generally several input pairs containing the same input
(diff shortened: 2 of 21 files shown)


clarity_ext/extensions.py
```diff
@@ -258,10 +260,11 @@ def _run(self, config, path, pid, module, artifacts_to_stdout, disable_context_c
                                           disable_commits=disable_context_commit,
                                           uploaded_to_stdout=artifacts_to_stdout)
         instance = extension(context)
-        context.logger.log("Start running script '{}'".format(module))
-        # We add the config in this way (not via __init__) as a quick fix.
-        instance.config = config
         try:
+            context.logger.log("Start running script '{}'".format(module))
+            context.validate_xml_concordance()
```
withrocks, line 265: How about just doing the validation when the data is required? I'm thinking e.g. pushing it into the `all_artifacts` method, or similar. That is, as close to the actual call to the genologics library.

I think that might make it easier to reason about and remove afterwards. A minor effect is that it could perform better, in the case we have an extension that doesn't require all artifacts.

clarity_ext/extensions.py
```diff
@@ -272,20 +275,29 @@ def _run(self, config, path, pid, module, artifacts_to_stdout, disable_context_c
...
-            # Ordinary files and objects shouldn't have been moved to upload queue,
-            # since the scipt have been interrupted by the usage error.
-            context.commit()
-            # UsageErrors are deferred and handled by the notify method
-            # To support the case (legacy) if someone raises an error without adding it to the defer list,
-            # we add it to the instance too:
-            if len(instance.errors) == 0:
-                instance.errors[e] = list()
+            self._commit_step_log(context, instance, e)
+        except Exception as e:
```
withrocks, line 280: If we grab exceptions at the lowest level (the `Exception` base class) I think we should always rethrow them. Otherwise we can have some very hard to debug issues later on.

clarity_ext/extensions.py
```diff
@@ -272,20 +275,29 @@ def _run(self, config, path, pid, module, artifacts_to_stdout, disable_context_c
             context.logger.log("Finished running script '{}'".format(module))
             context.commit()
         except UsageError as e:
-            # Commit in order to upload step log
-            # Ordinary files and objects shouldn't have been moved to upload queue,
-            # since the scipt have been interrupted by the usage error.
-            context.commit()
```
withrocks, line 278: I think this is a good change. Although the comment says that it's supposed to not have anything in the upload queue. I assume that might have been true at some point but then changed after some refactoring. Better to be sure and just commit the step log!

clarity_ext/repository/step_repository.py
```diff
@@ -44,16 +45,10 @@ def all_artifacts(self):
...
-
-        artifacts = self.session.api.get_batch(artifact_keys)
-        artifacts_by_uri = {artifact.uri: artifact for artifact in artifacts}
-        for input, output in input_output_maps:
-            input['uri'] = artifacts_by_uri[input['uri'].uri]
-            output['uri'] = artifacts_by_uri[output['uri'].uri]
+        wrappable_pairs = WrappablePairs(self.session, input_output_maps, WrappablePairs.MODE_STATELESS)
+        # We have to defer raising error until all artifacts are fetched.
+        # Step log is in the artifacts. The errors are to be written into the step log...
+        self.xml_discordance_errors = wrappable_pairs.validate()
```
withrocks, line 52: Related to: https://github.com/Molmed/clarity-ext/pull/209/files#r608413175

I would prefer if we would simply throw an error here if the data isn't valid. E.g. a `AssertionError`

withrocks: Maybe you don't do that because of that thing with `Step log` not being loaded yet? Correct me if I'm wrong :)

If that's the case: I don't think we need to worry about that because this error is not intended to be read by the end user but rather by us. We absolutely don't expect this to happen so in my mind it's enough that it just ends up in the system log. The step log was originally intended just for errors that are interesting to the user.

clarity_ext/repository/step_repository.py
```diff
@@ -67,26 +62,35 @@ def all_artifacts(self):
         # that we create only one artifact domain object in this case:
         outputs_by_id = dict()
         container_repo = ContainerRepository()
-        for input_res, output_res in input_output_maps:
+        for genologics_input, genologics_output, output_generation_type in wrappable_pairs:
```
withrocks: Could we rename `genologics_` to e.g. `rest_`, `api_` or `raw_`? I think we never refer to it as `genologics_` - probably just `api_`.

clarity_ext/repository/step_repository.py
```diff
@@ -169,3 +173,168 @@ def get_process_type(self):
...
+    Contains all artifact pairs within a step, containing genologics Artifact instances
+    These are in turn ready to be wrapped into clarity-ext domain instances
+    Usage:
+    wrappable_pairs = WrappablePairs(session, input_output_maps, mode=<stateful | stateless>)
+    wrappable_pairs.validate()  # compares xml for stateful and stateless
+    for input, output, output_generation_type in wrappable_pairs:
+        input  # genologics Artifact, either stateful or stateless
+        output  # genologics Artifact
+
+    The validation compares the stateless and statefull xml representation,
```
withrocks: One extra `l` in this `stateful` string :)

clarity_ext/repository/step_repository.py
```diff
@@ -169,3 +173,168 @@ def get_process_type(self):
...
+    and discards any discrepancies in the qc-flag
+    """
+
+    MODE_STATELESS = "stateless"
+    MODE_STATEFUL = "stateful"
+
+    def __init__(self, session, input_output_maps, state_mode):
+        # Note, the force flag should be set to False as soon as the test period is over!
```
withrocks, line 198: Cool :) I think we should even go further and just remove all the state-related code when it's over.

clarity_ext/repository/step_repository.py
```diff
@@ -169,3 +173,168 @@ def get_process_type(self):
...
+
+        return pairs
+
+    def _fetch_stateful(self):
+        artifact_keys = set()
+        for input, output in self.input_output_maps:
+            artifact_keys.add(input["uri"])
+            artifact_keys.add(output["uri"])
+
+        return self.session.api.get_batch(artifact_keys, force=self.force)
```
withrocks, line 261: This is a minor point, as it just concerns two calls. But can we fetch all in one batch call? I.e. merge `_fetch_stateful` and `_fetch_stateless`? Do you think that's unnecessary micro-optimization at this point?

clarity_ext/repository/step_repository.py
```diff
@@ -169,3 +173,168 @@ def get_process_type(self):
...
+        errors = list()
+        for pair in self.pairs:
+            errors.extend(pair.validate(self.session.api))
+        return errors
+
+    def __iter__(self):
+        self.counter = 0
+        return self
+
+    def next(self):
```
withrocks, line 220: I forget how this is with iteration classes. Are you supposed to implement `next`?

withrocks: It seems on this answer on SO that it's not required: https://stackoverflow.com/a/24377/282024

I might be missing something though :)

clarity_ext/repository/step_repository.py
```diff
@@ -169,3 +173,168 @@ def get_process_type(self):
...
+        return self.__next__()
+
+    def __next__(self):
+        if self.counter == len(self.pairs):
+            raise StopIteration
+        pair = self.pairs[self.counter]
+        self.counter += 1
+        return iter((pair.input_artifact, pair.output_artifact, pair.output_generation_type))
```
withrocks, line 228: Why is that `iter` keyword required?

clarity_ext/repository/step_repository.py
```diff
@@ -169,3 +173,168 @@ def get_process_type(self):
...
+        if self.state_mode == WrappablePairs.MODE_STATEFUL:
+            return artifact_repr.stateful_representation
+        elif self.state_mode == WrappablePairs.MODE_STATELESS:
+            return artifact_repr.stateless_representation
+        else:
+            raise Exception("Unknown state: {}".format(self.state_mode))
+
+    def validate(self, api):
+        errors = list()
+        if not api.is_equal(
```
withrocks: My feeling is that the check for equality should only be in clarity-ext. My reasoning is that we will eventually want to send all of our patches to upstream genologics, and in that case I would prefer that we don't need to argue for any of the patches. Since we're implementing this feature in clarity-ext, it makes sense to me that we also assert that it works as expected only in clarity-ext.

This is not saying that this might not be of use in genologics also, but it's just simpler to manage the future pull requests if we leave that question out of it :)

withrocks: ... but fetching an artifact with or without state seems basic enough to me for us to be able to get it into upstream eventually.

clarity_ext/service/file_service.py
```diff
@@ -221,22 +221,40 @@ def close_local_shared_files(self):
...
         for artifact_id in self.os_service.listdir(self.upload_queue_path):
             for file_name in self.os_service.listdir(os.path.join(self.upload_queue_path, artifact_id)):
-                if disable_commits:
-                    self.logger.info("Uploading (disabled) file: {}".format(os.path.abspath(file_name)))
+                associations.append((artifact_id, file_name))
+        return associations
+
+    def _commit_file(self, disable_commits, artifact_id, file_name):
+        """Copies files in the upload queue to the server"""
+        print('file name: {}'.format(file_name))
```
withrocks: `print` :)

clarity_ext/service/step_logger_service.py
```diff
@@ -29,6 +29,10 @@ def __init__(self, file_handle, file_service, raise_if_not_found=False, append=T
         # TODO: This should be configurable.
         self.NEW_LINE = "\r\n"
 
+    @property
+    def filename_w_extension(self):
```
withrocks: Should we maybe use `_with_` instead of `_w_`?

withrocks:
Looks cool! I would perhaps consider (I mentioned it in the genologics PR I think) to use the lxml diff library too: https://pypi.org/project/xmldiff/

Let me know your thoughts on my comments.

clarity_ext/extensions.py
GitEdvard: I had a bit of a head ache about this one. If I trigger the validation right away, the result of the validation can't be pushed to the step log, because the step log is fully initalized after all_artifacts().

clarity_ext/extensions.py
GitEdvard: :+1:

clarity_ext/extensions.py
GitEdvard: I fix this. The instance.error ... lines are actually included in _commit_step_log(), but in a quite hard to read way. So I'll change that.

clarity_ext/repository/step_repository.py
GitEdvard: Yes, that's right. I defer the raise error because of the step log thing. I wonder if it's ok to keep the writing to the step log? By writing to the step log, I got "hard proof" that an error was really registered when I validated. Furthermore, I only intend to keep this validation for a period of time, and then remove it (because all artifacts are loaded twice, and it ought to affect the performance).

clarity_ext/repository/step_repository.py
GitEdvard: Ok!

clarity_ext/repository/step_repository.py
GitEdvard: :+1:

clarity_ext/repository/step_repository.py
GitEdvard: :+1:

clarity_ext/repository/step_repository.py
GitEdvard: Because I want to decompose in a for loop, like "for input, output, generation_type in wrappable_pairs:"

clarity_ext/service/file_service.py
GitEdvard: ops!

clarity_ext/service/step_logger_service.py
GitEdvard: :+1:

clarity_ext/repository/step_repository.py
GitEdvard: ok!

clarity_ext/repository/step_repository.py
GitEdvard: After some thought, I wonder if I could leave it as is? I recall that this code was kind of "unstable". Because the input and output in input_output_maps mutates, and behaves a little different before and after an api fetch, I experienced a lot of unexpected behaviour from it. So, I'm afraid that this requires thorogh testing. 

I intend to have a period of a couple of months, when all artifact xml is validated on production in this way. If no error was encountered, only stateless artifacts will be fetched henceforth.

GitEdvard: I suspect there is some trickery with the cache and the state-flag. I would not be surprised if the state part is remove before adding it to the cache. (I had to add the force flag in order to fetch the artifacts a second time...)

clarity_ext/repository/step_repository.py
GitEdvard: is_equal read the xml, sort it, remove whitespaces to make it comparable with other instances. It also give the option to exclude a named tag (qc-flag). I didn't want to include it in clarity-ext because it's so very much about the internal of a genologics entity, (the xml). Apart from we are two groups, I think there's no doubt this function belongs to the genologics repo.

GitEdvard: Yes, I think I thought of this when I wrote the code, where to put it. I still think it belongs to the genologics repo, but if you have a strong opinion about this, I'll move it. I'm against the existence of a "genologics bubble" within clarity-ext. In the long run that could cause problems. I also think is_equal is rather innocent, because they simply don't reference it. Also, in case they want to merge with ours repo, wouldn't that be because they want our changes. I don't see that we have to argue about that?

GitEdvard (commented):
Thanks for the comments! I send in my replies now, and start doing adjustments. I have made a few replies that you might want to discuss further.

clarity_ext/repository/step_repository.py
GitEdvard: I believe next() is needed in python 2, but not in python 3

clarity_ext/repository/step_repository.py
GitEdvard: What about I rename it to a less generic name, like has_equal_xml() ??

clarity_ext/repository/step_repository.py
withrocks: Sounds good!

clarity_ext/repository/step_repository.py
withrocks: Yes, I was also wondering if this would be unnecessary complication :+1:

clarity_ext/repository/step_repository.py
withrocks: There are always politics in pull requests. I think they will want most of our patches, but in my experience (with any PR to public repos) is that they will also not want a more complex codebase unless they understand the need for it for the whole.

I'm just assuming that an equality check for the XML responses is not something they will see the need for. I imagine them asking why we can't just do that at another level. But I might be wrong!

An alternative that would also work is just to remove this equality check in genologics when we remove our use of it in clarity-ext. :thinking:

I'm fine with both approaches.

withrocks: If you use that xml diff engine (https://pypi.org/project/xmldiff/) the patch will be minimal and I think they might accept it (although then they might ask us why we're adding a new pip requirement for this :)).

withrocks: And I also want to just say that I also think that `is_equal` is as you say very innocent - and there are good arguments for including it. It's definitely not worth worrying too much about at least - we can always fix it if they don't want it.

GitEdvard: Ok, I'll move it to the clarity-ext. I also check out the xmldiff.

withrocks (approved):

PR #215, opened 2021-05-06
Python 3 sort key for container names

GitEdvard:
Purpose:
Fix a bug related to plate sorting in python 3

Background:
Previously, a sort key has been generated for each plate according to
ab1_plate3_210505 --> (\<container sort weight\>, \<is temp plate\>, 'ab', 1, 'plate', 3, 210505)

This approach do not longer work in python3, becaus (probably) there is control tubes in the list behind the scene, with name = lims-id.

sort key for blank control:
24-12345 --> (\<container sort weight\>, \<is temp plate\>, 24, 12345)

which will cause comparison between int and str. 

Solution:
The sort key has fixed places for ints and strings, where string are on even indexes and int is on odd indexes. 
ab1_plate_210505 --> (\<container sort weight\>, \<is temp plate\>, 'ab', 1, 'plate', 0, '', 210505)
24-12345 --> (\<container sort weight\>, \<is temp plate\>, '', 24, '', 12345)

note the "0" and the '', when there is no number or is no string for the current name part.

### DIFF
--- a/clarity_ext/service/dilution/service.py
+++ b/clarity_ext/service/dilution/service.py
@@ -602,15 +602,33 @@ class SortStrategy:
             # the list starting with an empty string should be sorted first into zip()
             return len(a_list[0])
 
-        strings_in_name = re.split('[-_]*\d+[-_]*', container.name)
-        strings_in_name = list(map(conv_lower, strings_in_name))
-        numbers_in_name = re.split('[-_]*\D+[-_]*', container.name)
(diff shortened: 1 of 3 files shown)


withrocks:
Looks great! I added some documentation, tests and moved a few lines of code, can you check it out?

https://github.com/Molmed/clarity-ext/pull/215/commits/289cca01dcd65d025cc20b04f0134e52c734cd1d

withrocks (approved):

GitEdvard:
Thanks for the documentation, and the improvements.

PR #216, opened 2021-05-07
WIP: Support forking extensions

withrocks:
Forked extensions run from one extension but use the same context as the
parent.

Example usage:

```
def execute(self):
    self.fork(SomeOtherExtension)
```

### DIFF
--- a/clarity_ext/extensions.py
+++ b/clarity_ext/extensions.py
@@ -226,7 +226,7 @@ class ExtensionService(object):
                 raise ValueError("Unexpected run argument type")
             return test
 
-        instance = self._get_extension(module)(None)
+        instance = self._get_extension(module)(None, None, None)
         ret = list(map(parse_run_argument, instance.integration_tests()))
         if require_tests and len(ret) == 0:
             raise NoTestsFoundException()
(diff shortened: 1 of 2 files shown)


clarity_ext/extensions.py
```diff
@@ -248,32 +248,18 @@ def _get_extension(self, module):
...
-                                          uploaded_to_stdout=artifacts_to_stdout)
-        instance = extension(context)
+    def run_instance(self, instance):
+        extension = instance.__class__
+        context = instance.context
         try:
-            context.logger.log("Start running script '{}'".format(module))
             context.validate_xml_concordance()
-            # We add the config in this way (not via __init__) as a quick fix.
-            instance.config = config
```
GitEdvard, line 268: Nice that you got this into init()

clarity_ext/extensions.py
```diff
@@ -226,7 +226,7 @@ def parse_run_argument(in_argument):
                 raise ValueError("Unexpected run argument type")
             return test
 
-        instance = self._get_extension(module)(None)
+        instance = self._get_extension(module)(None, None, None)
```
GitEdvard, line 229: This extra call to the extension has always confused me. It also makes it awkward to use the init() methods in extensions, if one should want to. Anyway, this is perhaps out of scope.

GitEdvard (approved):

PR #217, opened 2021-05-10
Allow config and extension_svc to be None

withrocks:
The arguments config and extension_svc need to be None so we don't have
to patch too much in scripts using this.


GitEdvard (approved):
Great!

PR #218, opened 2021-05-20
Use current state when fetching input artifacts

withrocks:
Depends on: https://github.com/Molmed/genologics/pull/22

It may help to diff this against the production branch: `git diff prod-at-snpseq -- $(git show --pretty= --name-only)`

---

When fetching a process via the API, the input output map has all
artifacts listed with a particular state. We've historically used
this state flag when updating an artifact.

However, there was one exception to this. QC flags where being fetched
using no `state` flag when they were being set. The script developer had
to force this with an API call that was only available in
clarity-snpseq, i.e. the scripts used at SNP&SEQ, Uppsala University.

This patch fixes this so we now fetch without that state parameter.

Note that we had already fixed this in an earlier patch (d03a3d6 and
related) but it was done in a different way and this patch does some
extra refactorings and fixes as there was a bug in the earlier patch.


clarity_ext/domain/aliquot.py
```diff
@@ -4,9 +4,9 @@
 
 class Aliquot(Artifact):
     """
-    NOTE: This class currently acts as a base class for both Analyte and ResultFile. It will be
-    merged with Analyte, since the name can cause some confusion as Analytes
-    and ResultFiles strictly don't need to be Aliquots, i.e. they can be non-divided copies
+    NOTE: This class currently acts as a base class for both Analyte and ResultFile. It should
+    be renamed (e.g. AnalyteBase), since the name can cause some confusion as Analytes
```
GitEdvard: So why don't you just change the name?

clarity_ext/extensions.py
```diff
@@ -252,7 +252,6 @@ def run_instance(self, instance):
         extension = instance.__class__
         context = instance.context
         try:
-            context.validate_xml_concordance()
```
GitEdvard, line 255: I don't get this. It was under your council that I implemented a concordance validation for xml's with and without state. Now you fetch artifacts without state as I did, and just remove the validation?

GitEdvard (commented):
I don't understand why you removed the xml concordance test. I spent some time doing that, simply because you thought it wise to do it.

GitEdvard:
On a second thought, I don't understand why this should fix the bug that I bumped into with the previous patch in ( in https://lims-dev.snpseq.medsci.uu.se/clarity/work-details/61886, although the qc flags have by now been set as they should). Both patches fetch artifacts with current state. To complicate things even more, Heidur detected a similar bug with qc flags not being set in aggregate in production, which completely uses lablogic to set qc flags. This could be that there is the same and only bug that shows up in both places (the new patch did not introduced a bug), or it could be two different bugs, with the new patch introducing another bug.

clarity_ext/domain/aliquot.py
withrocks: Good point :) I was just fixing this comment, because I thought the comment (which I think I wrote) was wrong. I.e. I thought this class shouldn't be merged with Analyte. But now when I'm thinking about it, should it maybe? I think it would be good to discuss this at some point. I'm often confused about the distinction between Analyte and Aliquot in clarity-ext.

I did however not want to change the name in this commit because we're just hotfixing. In any case, I think I'll revert to the old text.

withrocks:
About the XML check: It was because we now have it documented that fetching without the state flag just fetches the latest version. As I recall it, when we discussed this first we were under the assumption that the state flag wasn't documented.

I'm of course happy to put it back in if you think that it's important. I'm torn, but I think I lean towards that using a documented feature and doing a very thorough validation on staging is enough.

withrocks:
Regarding your comment on if this works or not. It actually seems to work. We tested all scenarios together (on my workstation), I and the REs.

The issue in production, I think that might also be because someone ran development code against production, but I haven't investigated that.

Now the reason for why it now works I think doesn't have to do with my refactoring (which happened because of the nature of how I worked on the bug). I think (but haven't checked it) that it simply has to do with that the QC flag wasn't being set on `Analytes`, but only on `ResultFiles`. I'll find the line and comment there!

clarity_ext/extensions.py
withrocks: It all had to do with that I first refactored the logic. With the refactored logic I was going to put this back in, but then I was confident in that `state` fetches the latest, since it's documented to do that.

I'm happy to put it back in if you think we should

GitEdvard (approved):
I merge this now, since there is a separate ticket for the xml concordance. Concerning why the validation case on dev wasn't working for me, you mention in the comment something with qc flag on result file. Is that fixed now?

PR #221, opened 2021-05-28
WIP: dev 1331, add sample repo 

GitEdvard:
This is WIP because it should not be included into the current release (2021-05-28)

Purpose:
Adjust clarity-ext so that one can fetch related data (i.e. samples) from container name as the primary search key. This is to be used by the snpseq-data repository

Implementation:
Add a sample respository. This is initalized during setup with all samples to be fetched from the current context, which can be either a step or a container. At first call to get_samples(), all samples are fetched at one swoop.

In addition, I also implemented an inversion of control instance, in order to avoid sending in each and every repository as input argument to domain classes during initialization.

### DIFF
--- a/clarity_ext/domain/aliquot.py
+++ b/clarity_ext/domain/aliquot.py
@@ -15,17 +15,13 @@ class Aliquot(Artifact):
     QC_FLAG_UNKNOWN = 'UNKNOWN'
 
     def __init__(self, api_resource, is_input, id=None, samples=None, name=None,
-                 well=None, qc_flag=None, udf_map=None, mapper=None):
+                 well=None, qc_flag=None, udf_map=None, mapper=None, sample_repo=None):
         super(Aliquot, self).__init__(api_resource=api_resource,
                                       artifact_id=id,
                                       name=name,
                                       udf_map=udf_map,
                                       is_input=is_input,
-                                      mapper=mapper)
-        # NOTE: This is a quick fix for extremely slow loading of large pools
-        if samples:
-            self._samples_require_initializing = not isinstance(samples[0], Sample)
-        self._samples = samples
+                                      mapper=mapper,)
 
         self.well = well
         if well:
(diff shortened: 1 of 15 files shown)


clarity_ext/domain/aliquot.py
```diff
@@ -21,11 +22,7 @@ def __init__(self, api_resource, is_input, id=None, samples=None, name=None,
                                       name=name,
                                       udf_map=udf_map,
                                       is_input=is_input,
-                                      mapper=mapper)
-        # NOTE: This is a quick fix for extremely slow loading of large pools
-        if samples:
-            self._samples_require_initializing = not isinstance(samples[0], Sample)
-        self._samples = samples
+                                      mapper=mapper,)
```
withrocks, line 25: Is that extra comma something the IDE is adding?

clarity_ext/mappers/clarity_mapper.py
```diff
@@ -47,6 +48,8 @@ def _get_from_cache(self, domain_object):
         return self.map[domain_object]
 
     def sample_create_object(self, resource):
+        # from clarity_ext.utils import pretty_print
+        # pretty_print(resource)
```
withrocks, line 52: Extra comments.

clarity_ext/domain/aliquot.py
GitEdvard: I have a feeling that's me adding and removing lines :/

clarity_ext/domain/container.py
```diff
@@ -210,6 +210,19 @@ def columns(self):
...
+    @property
+    def samples(self):
+        """
+        List an unique set of samples in a container. Note that the same control sample
+        may occur in different pools, are here filtered.
+        :return:
+        """
+        analytes = [well.artifact for well in self.occupied]
+        sample_dict = {s.id: s for a in analytes for s in a.samples}
+        samples = [v for _, v in sample_dict.items()]
```
withrocks, line 221: FYI: You can also do `.values()`

withrocks:
Cool!

withrocks (approved):

PR #223, opened 2021-06-16
lims-2324, Handle encoding for empty files

GitEdvard:
This is a followup bugfix. Handle empty files when asserting which encoding a file has. This happens for the empty step logs when entering a step.

### DIFF
--- a/clarity_ext/repository/file_repository.py
+++ b/clarity_ext/repository/file_repository.py
@@ -28,9 +28,9 @@ class FileRepository:
         with open(local_path, 'rb') as f:
             byte_contents = f.read()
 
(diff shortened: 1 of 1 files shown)


clarity_ext/repository/file_repository.py
```diff
@@ -29,6 +29,8 @@ def open_local_file(self, local_path, mode):
             byte_contents = f.read()
 
         dammit = UnicodeDammit(byte_contents, ['uft-8', 'latin-1'])
```
withrocks: `uft` => `utf`

withrocks (approved):

PR #224, opened 2021-06-17
Templates should use the newline function from the base class

withrocks:
After moving to Py3 line endings are automatically changed to
'\n', even if they use '\r\n'. So Windows line endings where
being removed.

This solution requires the user to specify what line ending is used
rather than using the one that's in the template file.


GitEdvard:
OK. I merge it although the checks haven't completed. It seems something is wrong with them.

PR #226, opened 2021-06-29
Dilution slow

withrocks:
A few changes related to the slowness of the dilution scripts.

The main change is that we now compare objects simply by `id` and only by supporting the `__lt__` method.

There are other changes that are related to pytest warnings and `DomainObjectMixin` and `DomainObjectWithUdfMixin` are renamed so they don't have `Mixin` in their name. This is because they are only ever used as base classes.

The auth monkey patch is being removed from the frozen tests feature as we don't use it anymore and some tests were failing because of it.

### DIFF
--- a/clarity_ext/domain/common.py
+++ b/clarity_ext/domain/common.py
@@ -10,29 +10,6 @@ class DomainObjectMixin(object):
         else:
             return False
 
-    def __hash__(self):
-        return hash(self.id)
-
-    def __lt__(self, other):
-        return self.compare(other) < 0
-
-    def __gt__(self, other):
-        return self.compare(other) > 0
-
-    def __le__(self, other):
-        return self.compare(other) <= 0
-
-    def __ge__(self, other):
-        return self.compare(other) >= 0
-
-    def compare(self, other):
-        # Override if needed
-        if self._eq_rec(self, other) == 0:
-            return 0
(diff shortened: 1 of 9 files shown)


clarity_ext/domain/common.py
```diff
@@ -2,36 +2,18 @@
...
-        return self.compare(other) >= 0
-
-    def compare(self, other):
-        # Override if needed
-        if self._eq_rec(self, other) == 0:
-            return 0
-        elif str(self) < str(other):
-            return -1
-        return 1
+        return self.id < other.id
```
GitEdvard, line 16: Is `__lt__` tested for when a list contains None values?

GitEdvard (commented):
Great work. I think that we should add a ticket to change behaviour of equality checks, making use of id's only. For what I can see here, however , the `__lt__` will cast exception for lists containing None's.

clarity_ext/domain/common.py
withrocks: Excellent catch :)

withrocks: I think that it's perhaps OK if one gets an error if trying to sort an object with incompatible objects or None. One can always fix it by explicitly providing a key function:

```sorted(lst, key=lambda a: a.id if a else 0)```

PR #230, opened 2021-09-06
develop-1565, make refactorings in test framework for showing proper error message

GitEdvard:
Purpose:
Make some adjustment to the testing framework to also include tests for the holding step.

Details:
The extension_builder_base initializes udf values in two steps
1) Register that an analyte (that is all analytes in a step) is configured with an udf with a specific name. That means that a script may populate this udf during execution
2) Pre-configure a specific analyte in a step with a value

For this, I implemented that the pair builder may inherit a default pair builder. The default pair builder is initialized according point 1) above

Note, the "read_result_file_builder.py" was renamed to "extension_builder_base.py", and also modified. But "extension_builder_base.py" is shown as a complete new file in the change log.


matrulda (approved):
Looks good! 👍

PR #233, opened 2022-03-30
Fix a bug that happens at step commit

GitEdvard:
Description:
The bug is related to the domain object instantiation process in clarity-ext. In clarity, analytes can be fetched at high speed, because they are included in the api response for the step. The related entity "sample" (think original sample), on the other side, takes long time to fetch. For a while ago, a feature was introduced, that fetched all samples in a step in one swoop. When the first sample was requested, instead of fetching one single sample, all samples were fetched for the current step. 

The proper instantiation goes like this:
1. fetch xml representation for an object (analyte or sample), instantiate a python domain object
2. Register this object in clarity_mapper, which contains a cache (evil?!) of all objects in the step
3. Script execution updates the domain object
4. Push up the updated domain object to API in commit method. Here, the object is sometimes once again fetched from the cache in clarity_mapper.

The bug here was that clarity_mapper was instanciated locally in sample_repository. It needs to be the very same instance that is used at domain instantiation (step 2 above), as in the commit action (step 4 above). 

Validation:
* All unit tests pass
* Integration test for qc.qc_commit.py runs without failure

### DIFF
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -77,8 +77,8 @@ class ExtensionContext(object):
         use the constructor for custom use and unit tests.
         """
         session = ClaritySession.create(step_id)
-        ioc.set_application(ApplicationService(session))
         clarity_mapper = ClarityMapper()
+        ioc.set_application(ApplicationService(session, clarity_mapper))
         step_repo = StepRepository(session, clarity_mapper)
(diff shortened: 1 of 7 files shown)


matrulda (approved):
Looks good to me! Thanks for the detailed description. 

Is it expected that travis fails?

PR #234, opened 2022-05-03
Remove old tweak code

GitEdvard:
There was a conditional statement in is_pool() method, Aliquot class, that was present only to
fix unit tests. Since the handling of samples were changed some time
ago, this is both unnecessary (the test pass anyway), and induce a bug.
So, remove this conditional statement fixes the whole issue.


kleppari (approved):

kleppari:
Looks good

PR #235, opened 2022-05-03
lims 2123, Fix cicd

GitEdvard:
Purpose: 
The CI/CD stopped to work under travis. 

Solution:
I switched from travis to github actions instead. 

Validation:
I have verified that clarity-ext is installed without errors (in logs). I have also pushed a failing unit test and verified that the CI/CD fails in that case.


.github/workflows/github-actions-demo.yml
```diff
@@ -0,0 +1,24 @@
+name: GitHub Actions Demo
```
matrulda: It would be nice to rename this file, workflow and job name to something similar to "Run clarity-ext unit tests"

.github/workflows/github-actions-demo.yml
```diff
@@ -0,0 +1,24 @@
...
+on: [push, pull_request]
+jobs:
+  Explore-GitHub-Actions:
```
matrulda: See above comment.

.github/workflows/github-actions-demo.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
```
matrulda: There's a risk something could break when just using latest, so I would suggest using a specific version, for example ubuntu-20.04

.github/workflows/github-actions-demo.yml
```diff
@@ -0,0 +1,24 @@
...
+    strategy:
+      matrix:
+        python-version: [3.7]
+
+    steps:
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
```
matrulda: I happen to know there's a v3 for this release action, perhaps use that instead? When implementing something new I think it's worth trying to use the most recent versions :)

matrulda (changes requested):
Looks good! Left some minor comments. Not really necessary, but I think it would be nice to change the demo names at least. Let me know if you feel differently. 

I'm not sure while one of the jobs still are running? But hopefully it's just a temporary hickup.

.github/workflows/github-actions-demo.yml
GitEdvard: Yes, when I arrived at home yesterday, I realized that I forgot to rename the demo file :)

.github/workflows/github-actions-demo.yml
GitEdvard: :+1:

.github/workflows/github-actions-demo.yml
GitEdvard: :+1:

GitEdvard:
Thank you for the feedback. I have now adjusted accordingly.

matrulda (approved):
Nice, good job! Great to get rid of travis :smile:

PR #236, opened 2022-05-12
Allow for staged fetching in sample repository

GitEdvard:
Bug description:
Recently, sample are fetched by the new Sample repository class. Here, all samples within one step are fetched in one swoop, instead of one by one, which will increase the performance. This approach didn't work out in "SNP&SEQ Denature (NovaSeq) v1" where samples are fetched from more than one step. 

Sample repo implementation background:
Fetching samples goes in two steps. 1) register a sample as a candidate. Only the sample uri is cached in this step. The sample uri is already contained within analytes by defaults. 2) fetch all sample information by a batch api call. 

Bug fix:
Update sample repository class, so that it's possible to to fetch samples in several occasions (in this example, fetch samples for each parent step that occured) . Before, it was only permitted once.

### DIFF
--- a/clarity_ext/repository/sample_repository.py
+++ b/clarity_ext/repository/sample_repository.py
@@ -14,7 +14,6 @@ class SampleRepository(object):
         self.clarity_mapper = clarity_mapper
         self.candidates = list()  # not-fetched genologics sample resources
         self.samples = dict()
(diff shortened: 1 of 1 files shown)


kleppari (approved):
Looks good

PR #241, opened 2023-03-03
Update genologics submodule to version with special character fix

slohse:
Getting the [special character fix](https://github.com/Molmed/genologics/pull/26) upstream...


matrulda (approved):
:+1:

PR #242, opened 2023-03-14
Added support for xlsx files

kleppari:
- Support for reading xlsx files using openpyxl

### DIFF
--- a/clarity_ext/context.py
+++ b/clarity_ext/context.py
@@ -221,7 +221,8 @@ class ExtensionContext(object):
         self.validation_service.handle_validation(
             self.validation_results, tailored_error_message=msg)
 
-    def local_shared_file(self, clarity_file_handle, mode="r", is_xml=False, is_csv=False, file_name_contains=None):
+    def local_shared_file(self, clarity_file_handle, mode="r", is_xml=False, is_csv=False, file_name_contains=None,
+                          is_xlsx=False):
         """
         Downloads the file from the current step. The returned file is generally a regular
         file-like object, but can be casted to an xml object or csv by passing in is_xml or is_csv.
@@ -246,6 +247,9 @@ class ExtensionContext(object):
         elif is_csv:
             check_file_extension('.csv')
             return self.file_service.parse_csv(f)
+        elif is_xlsx:
+            check_file_extension('.xlsx')
+            return self.file_service.parse_xlsx(f)
         else:
             return f
 
--- a/clarity_ext/service/file_service.py
+++ b/clarity_ext/service/file_service.py
@@ -7,6 +7,7 @@ from zipfile import ZipFile
 from lxml import objectify
 from clarity_ext import utils
 import requests
+from openpyxl import load_workbook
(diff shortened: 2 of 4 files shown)


clarity_ext/context.py
```diff
@@ -221,7 +221,8 @@ def handle_validation(self, msg=None):
         self.validation_service.handle_validation(
             self.validation_results, tailored_error_message=msg)
 
-    def local_shared_file(self, clarity_file_handle, mode="r", is_xml=False, is_csv=False, file_name_contains=None):
+    def local_shared_file(self, clarity_file_handle, mode="r", is_xml=False, is_csv=False, file_name_contains=None,
```
Aratz, line 224: I know the body of this function is old code, but wouldn't it be better to remove the extra `is_*` arguments and only rely on detecting the file type automatically? It seems the bulk of what this function is doing is handling the case where the flag and the file type don't agree.

So I'd suggest either remove all the flags and detect the file type by looking at the extension in the file name, or, if it's important to have the user specify the file type they expect, to replace the flag args by `extension` where one could pass a string e.g. 'csv'

clarity_ext/context.py
```diff
@@ -246,6 +247,9 @@ def check_file_extension(extension):
         elif is_csv:
             check_file_extension('.csv')
             return self.file_service.parse_csv(f)
+        elif is_xlsx:
+            check_file_extension('.xlsx')
```
Aratz, line 251: Can it happen that the extension of the sample form is just `.xls`?

clarity_ext/service/file_service.py
```diff
@@ -58,6 +58,16 @@ def __init__(self, artifact_service, file_repo, should_cache, os_service, upload
...
+    def parse_xlsx(self, f):
+        from openpyxl import load_workbook
+        import pandas as pd
+        with open(f.name, "rb") as wbook:
+            writer = pd.ExcelWriter(wbook, engine='openpyxl')
+            workbook = load_workbook(wbook, data_only=True)
+
+            writer.book = workbook
+            return writer.book
+
```
Aratz: I tried this locally but I get an `AttributeError: can't set attribute` when reaching `writer.book = workbook` :confused: 

Is `pd.ExcelWriter` necessary here or can you return `workbook` immediately?

Aratz (changes requested):
Looks good!

My main question is regarding the `local_shared_file` function, it feels like maybe it could benefit from a rewrite.

The other main question is with the `parse_xlsx` function, I tried it locally but couldn't really make it work.

clarity_ext/context.py
kleppari: I will create a ticket for this

clarity_ext/context.py
kleppari: No, the sample from is just .xlsx

Aratz (approved):
Looks good, although I think you need to add `openpyxl` to the requirements for the tests to pass.

